PR #6, opened 2024-06-27
#Issue 1 -  Parse input configuration

Guillaume-Villain:
- Add parsing of the federation controller with the flag module
- Structure of the YAML is put into internal/pkg/config as exportable
- a global variable is created containg the yaml of the command line


cmd/federation-controller/main.go
jewertow, line 1: What do you think about the following ordering of variables and functions?
```
import (
	"flag"
	...
)

var (
	meshPeer = flag.String("meshPeers", "meshPeers Yml",
	exportedServiceSet = flag.String("exportedServiceSet", "exportedServiceSet Yml",
	importedServiceSet = flag.String("importedServiceSet", "importedServiceSet Yml",
)

func main() {
        flag.Parse()
	appConfig := parse()
```

jewertow (commented):
Thanks, I verified that parsing works correctly. I left a few suggestions.

cmd/federation-controller/main.go
```diff
@@ -1,7 +1,55 @@
...
+	"flag"
+	"fmt"
+	"os"
+	"github.com/jewertow/federation/internal/pkg/config"
+	"gopkg.in/yaml.v3"
+)
+
+// Global variable to store the parsed arguments and "flag" arguments
+var (
+	AppConfig config.Federation
```
jewertow, line 13: We shouldn't export this variable, because we will instantiate dependent components in this file, like exported service set watcher, and then we won't be able to import this config variable, because that would cause circular dependency.
```suggestion
```

cmd/federation-controller/main.go
```diff
@@ -1,7 +1,55 @@
...
+	unmarshalYAML(*meshPeer, &peers)
+	unmarshalYAML(*exportedServiceSet, &exported)
+	unmarshalYAML(*importedServiceSet, &imported)
+
+	AppConfig = config.Federation{
+		MeshPeers:          peers,
+		ExportedServiceSet: exported,
+		ImportedServiceSet: imported,
+	}
+}
```
jewertow, line 49: Regarding first comment:
```suggestion
func parse() *config.Federation {
	var (
		peers    config.MeshPeers
		exported config.ExportedServices
		imported config.ImportedServices
	)

	unmarshalYAML(*meshPeer, &peers)
	unmarshalYAML(*exportedServiceSet, &exported)
	unmarshalYAML(*importedServiceSet, &imported)

	return &config.Federation{
		MeshPeers:          peers,
		ExportedServiceSet: exported,
		ImportedServiceSet: imported,
	}
}
```

jewertow (changes requested):
Could you use snake case for file names?

cmd/federation-controller/main.go
Guillaume-Villain: Remove from export

cmd/federation-controller/main.go
Guillaume-Villain: change done

jewertow (approved):
Thanks

PR #7, opened 2024-07-12
Xds input

Guillaume-Villain:
Add the xds server that send to subscriber "test ServiceEntry"


internal/pkg/xds/xds_mcp.go
```diff
@@ -0,0 +1,202 @@
...
+	}
+
+	mcpResBody := &anypb.Any{}
+	if err := anypb.MarshalFrom(mcpResBody, seSpec, proto.MarshalOptions{}); err != nil {
+		return nil, fmt.Errorf("serializing ServiceEntry to protobuf message: %w", err)
+	}
+
+	mcpResTyped := &mcpv1alpha1.Resource{
+		Metadata: &mcpv1alpha1.Metadata{
+			Name: fmt.Sprintf("istio-system/mcp-example-%03d", idx),
```
jewertow, line 191: If we return fake data, could we return working SE? Then I would be able to install Istio with this controller and test provided resources.
```suggestion
func makeMCPServiceEntry(hostname string) (*anypb.Any, error) {
	seSpec := &istionetv1alpha3.ServiceEntry{
		Hosts:    []string{hostname},
		Location: istionetv1alpha3.ServiceEntry_MESH_EXTERNAL,
		Ports: []*istionetv1alpha3.ServicePort{{
			Number:   443,
			Name:     "https",
			Protocol: "TLS",
		}},
		Resolution: istionetv1alpha3.ServiceEntry_DNS,
	}

	mcpResBody := &anypb.Any{}
	if err := anypb.MarshalFrom(mcpResBody, seSpec, proto.MarshalOptions{}); err != nil {
		return nil, fmt.Errorf("serializing ServiceEntry to protobuf message: %w", err)
	}

	mcpResTyped := &mcpv1alpha1.Resource{
		Metadata: &mcpv1alpha1.Metadata{
			Name: fmt.Sprintf("istio-system/se-%s", hostname),
```

internal/pkg/xds/xds_mcp.go
```diff
@@ -0,0 +1,202 @@
...
+		Nonce: version,
+	}); err != nil {
+		return err
+	}
+	return nil
+}
+
+// pushToSubscribers pushes MCP resources to active subscribers.
+func (adss *adsServer) pushToSubscribers() error {
+	mcpResources, err := makeMCPResources(numMCPResources)
```
jewertow, line 111: I would like to provide some working test data. Could we send the following service entries?
```suggestion
var mcpResources []*anypb.Any
example, err := makeMCPServiceEntry("example.com")
mcpResources = append(mcpResources, example)
istio, err := makeMCPServiceEntry("istio.io")
mcpResources = append(mcpResources, istio)
```

PR #9, opened 2024-07-17
Add service import function

amadeus-mkandula:
Added a function to create a ServiceEntry or WorkloadEntry as appropriate for imported remote services.

### DIFF
--- a/internal/pkg/config/import_services.go
+++ b/internal/pkg/config/import_services.go
@@ -1,5 +1,16 @@
 package config
 
-type ImportedServices struct {
+import (
+	istionetv1alpha3 "istio.io/api/networking/v1alpha3"
+)
+
+type ImportedServiceSet struct {
 	Rules []Rules `yaml:"rules"`
+}
+
+type ImportedService struct {
+	AppName         string
+	ServiceHostname string
+	ServicePorts    []*istionetv1alpha3.ServicePort
+	Endpoints       []Remote
 }
--- a/internal/pkg/mcp/import_watcher.go
+++ b/internal/pkg/mcp/import_watcher.go
@@ -1,0 +1,109 @@
+package mcp
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/types/known/anypb"
+	api_v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	mcpv1alpha1 "istio.io/api/mcp/v1alpha1"
+	istionetv1alpha3 "istio.io/api/networking/v1alpha3"
+)
+
+// MakeImportedMCPResources creates a ServiceEntry or WorkloadEntry
+// for an imported remote service as needed.
+func MakeImportedMCPResources(serviceHostname string, appName string, peerName string, peerIngressEps *api_v1.EndpointSubset, peerLocality string, peerNetwork string, servicePorts []*istionetv1alpha3.ServicePort, serviceController *Controller) (*anypb.Any, error) {
+	serviceDnsNameSplit := strings.Split(serviceHostname, ".")
+	serviceName := serviceDnsNameSplit[0]
+	serviceNamespace := serviceDnsNameSplit[1]
+
+	ingressPortMap := makeIngressPortMap(servicePorts)
+	workloadEntries := makeWorkloadEntries(appName, peerIngressEps, peerLocality, peerNetwork, ingressPortMap)
+
+	var mcpResObjectName string
+	var mcpResBody *anypb.Any
+	service, err := serviceController.clientset.CoreV1().Services(serviceNamespace).Get(context.TODO(), serviceName, v1.GetOptions{})
+	if err != nil {
+		if errors.IsNotFound(err) {
+			// User created service doesn't exist, create ServiceEntry.
+			seSpec := &istionetv1alpha3.ServiceEntry{
+				Hosts:      []string{serviceHostname},
+				Ports:      servicePorts,
+				Endpoints:  workloadEntries,
+				Location:   istionetv1alpha3.ServiceEntry_MESH_INTERNAL,
+				Resolution: istionetv1alpha3.ServiceEntry_STATIC,
+			}
+
+			if err := anypb.MarshalFrom(mcpResBody, seSpec, proto.MarshalOptions{}); err != nil {
+				return nil, fmt.Errorf("Serializing ServiceEntry to protobuf message: %w", err)
+			}
+			mcpResObjectName = fmt.Sprintf("istio-system/import-%s", serviceName)
+		} else {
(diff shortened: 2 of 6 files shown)


internal/pkg/mcp/controller.go
```diff
@@ -0,0 +1,203 @@
...
+	informer     cache.SharedIndexInformer
+	eventHandler Handler
+	resourceType string
+}
+
+func objName(obj interface{}) string {
+	return reflect.TypeOf(obj).Name()
+}
+
+func StartControllers(ctx context.Context) {
```
jewertow, line 51: In general, we will not need informers to handle imported service, because they will come from the remote server and won't be created as local `Service` objects.

matteo-casalino: the idea is just to rely on the informer as a cache for local `Service`s, which we query to know whether we need to create a `ServiceEntry` or not.

jewertow: Ah right!

internal/pkg/mcp/utils.go
```diff
@@ -0,0 +1,45 @@
...
+func buildOutOfClusterConfig() (*rest.Config, error) {
+	kubeconfigPath := os.Getenv("KUBECONFIG")
+	if kubeconfigPath == "" {
+		kubeconfigPath = os.Getenv("HOME") + "/.kube/config"
+	}
+	return clientcmd.BuildConfigFromFlags("", kubeconfigPath)
+}
+
+// GetClientOutOfCluster returns a k8s clientset to the request from outside of cluster
+func GetClientOutOfCluster() kubernetes.Interface {
```
jewertow, line 36: We will never run federation controller out of cluster, so you can remove it.

jewertow:
@amadeus-mkandula can you rebase your branch?

amadeus-mkandula:
Rebased and removed my service controller code in favour of @matteo-casalino 's

internal/pkg/config/import_services.go
```diff
@@ -1,5 +1,16 @@
...
+	istionetv1alpha3 "istio.io/api/networking/v1alpha3"
+)
+
+type ImportedServiceSet struct {
 	Rules []Rules `yaml:"rules"`
 }
+
+type ImportedService struct {
+	AppName         string
+	ServiceHostname string
```
jewertow, line 13: Wouldn't be service name and namespace enough? I think even if the source cluster uses custom cluster domain, i.e. other than `cluster.local`, we do not care about that, because we will use east-west gateway's host name and SNI for routing. So to avoid splitting this field I suggest the following change:
```suggestion
	Name string
        Namespace string
```

internal/pkg/config/import_services.go
```diff
@@ -1,5 +1,16 @@
...
+	ServicePorts    []*istionetv1alpha3.ServicePort
```
jewertow, line 14: I wonder how we can obtain Istio ports if we watch k8s services...? It seems that we will have to require specifying `appProtocol` in exported services. Only then we are able to translate it to proper Istio protocol.

internal/pkg/config/import_services.go
```diff
@@ -1,5 +1,16 @@
...
+	Endpoints       []Remote
```
jewertow, line 15: Currently, we allow to specify single remote peer in the input configuration, so this will never be an array of remote peers. However, I agree that we want to support more than one remote peers, so we should also change the configuration structure.

jewertow: Beside that, I would prefer decouple imported service info and remote peer specification. In my opinion, `MakeImportedMCPResources` should take imported service and remote peer spec as separate arguments.

jewertow (approved):
I left a few comments. At this stage of the project we can merge it as is and implement improvements in follow-up PRs. It's up to you @amadeus-mkandula.

jewertow:
@matteo-casalino can you share your feedback?

internal/pkg/config/import_services.go
amadeus-mkandula: Yes that's true indeed, I hadn't considered that the local and remote clusters might be using different cluster domains so it doesn't make sense to import the full hostname.

If I understood correctly we will still need to build the expected internal cluster hostname when we create a ServiceEntry but we should be able to build this with the imported information + retrieving the local cluster domain. For example if we import remote service httpbin in the httpbin namespace we will have something like:

```
apiVersion: networking.istio.io/v1alpha3
kind: ServiceEntry
metadata:
  name: import-httpbin
  namespace: istio-system
spec:
  hosts:
  - httpbin.httpbin.svc.cluster.local
  ports:
  - number: 443
    name: https
    protocol: HTTPS
  endpoints:
  - address: remote.gateway.example.com
    ports:
      https: 15443
```

internal/pkg/config/import_services.go
amadeus-mkandula: Indeed I was thinking we would likely need to export the port protocol from the remote service. Either that or we could perhaps use some similar logic that Istio uses to determine the port protocol from the port name: https://github.com/istio/istio/blob/master/pkg/config/kube/conversion.go#L47-L87

But probably better still to explicitly export the port.

internal/pkg/config/import_services.go
amadeus-mkandula: Ah yes I see, so I suppose we would instead just keep a list of remote names that expose the service in the service spec and retrieve the remote endpoint spec from the second input as appropriate?

internal/pkg/mcp/import_watcher.go
```diff
@@ -0,0 +1,123 @@
...
+		}
+		for _, addr := range endpoint.Addresses {
+			ingressPortMap := makeIngressPortMap(importedService.ServicePorts, endpoint.Ports.DataPlane)
+			entries = append(entries, &istionetv1alpha3.WorkloadEntry{
+				Address:  addr,
+				Ports:    ingressPortMap,
+				Network:  endpoint.Network,
+				Locality: endpoint.Locality,
+				Labels: map[string]string{
+					"app":                       importedService.AppName,
```
matteo-casalino, line 90: Eventually I guess we'll want to support arbitrary selectors. For now using `app` is ok to have the MVP working. ?Maybe we should rather use a more recent standard such as `app.kubernetes.io/name`?

internal/pkg/mcp/import_watcher.go
```diff
@@ -0,0 +1,123 @@
...
+
+	for _, importedService := range importedServices {
+		serviceDnsNameSplit := strings.Split(importedService.ServiceHostname, ".")
+		serviceName := serviceDnsNameSplit[0]
+		serviceNamespace := serviceDnsNameSplit[1]
+
+		workloadEntries := makeWorkloadEntries(importedService)
+
+		mcpResBody := &anypb.Any{}
+		service, err := serviceController.clientset.CoreV1().Services(serviceNamespace).Get(context.TODO(), serviceName, v1.GetOptions{})
```
matteo-casalino, line 32:
```suggestion
		service, err := serviceController.informer.GetIndexer().GetByKey(cache.NewObjectName(serviceNamespace, serviceName).String())
```

Issue #2, opened 2024-06-26
Implement exported service watcher


jewertow:
Exported service watcher should watch k8s Services matching specified label selectors and trigger MCP update on create/delete events. For instance, a create event should result in sending updated Gateway resource with new Service DNS name appended to the `spec.servers[0].hosts`. The watcher should send new MCP resources to the XDS server via a channel or calling proper method, but should not handle XDS protocol itself.

matteo-casalino:
I should have some time to take this one in the coming days, unless already taken?

PR #10, opened 2024-07-19
issue #2. Introduce generic k8s controller based on informers and rat…

matteo-casalino:
…e-lmiting queue for debouncing and use it to implement export controller.


internal/pkg/mcp/export_handler.go
```diff
@@ -0,0 +1,89 @@
...
+	klog.Infof("Updated service %s, namespace %s", oldService.Name, oldService.Namespace)
+	w.pushMCPUpdateIfMatchesRules([]*corev1.Service{oldService, newService})
+}
+
+func (w *ExportedServiceSetHandler) pushMCPUpdateIfMatchesRules(services []*corev1.Service) {
+	for _, rules := range w.cfg.ExportedServiceSet.Rules {
+		if rules.Type != "LabelSelector" {
+			continue
+		}
+		for _, selectors := range rules.LabelSelectors {
```
JulienDelacroix, line 61:
```		for _, selectors := range rules.LabelSelectors {
			for _, service := range services {
```
I think that these 2 loops should be moved inside `matchesLabelSelector`, to make it return true if  any selector matches any service, to generate the Gateway once.
Otherwise you will re-generate the same Gateway twice when both old and new service matches, or even more when service matches multiple selectors...

internal/pkg/mcp/export_handler.go
```diff
@@ -0,0 +1,89 @@
...
+						w.pushMCP <- gateway
+					}
+					return
+				}
+			}
+		}
+	}
+}
+
+func matchesLabelSelector(obj *corev1.Service, matchLabels map[string]string) bool {
```
JulienDelacroix, line 77: I think this should take both old and new service + a list of selectors (see my comment in `pushMCPUpdateIfMatchesRule`)

internal/pkg/mcp/export_handler.go
matteo-casalino: in fact we exit the loop as soon as we find at least one matching service (either the old or new one). `generateGatewayForExportedServices()` then relies on the informer cache for generating only one `Gateway` for all matching services.

jewertow (approved):
Looks good and works fine. Can you change target branch to master?

internal/pkg/xds/xds_mcp.go
```diff
@@ -130,81 +128,81 @@ func (adss *adsServer) push(resources []*anypb.Any) error {
...
+// 			klog.Errorf("Error sending MCP resources: %v", err)
+// 			value.(*subscriber).closeStream()
+// 			adss.subscribers.Delete(key)
+// 		}
+
+// 		return true
+// 	})
+
+// 	return nil
+// }
```
jewertow, line 161: @matteo-casalino can you remove commented code? We can always look at previous commits or pull requests to reuse old code.

PR #23, opened 2024-08-13
WIP: #Issue 16 - Replace `klog` and `fmt.Println` with `istio.io/istio/pkg/log` 

yxun:
This PR is a change according to  https://github.com/jewertow/federation/issues/16
It replaces `klog` and `fmt.Print` with `istio.io/istio/pkg/log` 

This PR is adding the `spf13/cobra` command for parsing flags in the `cmd/federation-controller/main.go`

Question and Discussion are in  https://github.com/jewertow/federation/issues/16

### DIFF
--- a/cmd/federation-controller/main.go
+++ b/cmd/federation-controller/main.go
@@ -5,35 +5,63 @@ import (
 	"encoding/json"
 	"flag"
 	"fmt"
-	discovery "github.com/envoyproxy/go-control-plane/envoy/service/discovery/v3"
-	"github.com/jewertow/federation/internal/pkg/federation"
-	"github.com/jewertow/federation/internal/pkg/xds/adsc"
-	"log"
 	"os"
 	"os/signal"
 	"sync"
 	"syscall"
 
+	discovery "github.com/envoyproxy/go-control-plane/envoy/service/discovery/v3"
+	"github.com/jewertow/federation/internal/pkg/federation"
+	"github.com/jewertow/federation/internal/pkg/xds/adsc"
+
 	"github.com/jewertow/federation/internal/pkg/config"
 	"github.com/jewertow/federation/internal/pkg/mcp"
 	"github.com/jewertow/federation/internal/pkg/xds"
 	"github.com/jewertow/federation/internal/pkg/xds/adss"
+	"github.com/spf13/cobra"
+	istiolog "istio.io/istio/pkg/log"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
-	"k8s.io/klog/v2"
 )
 
-// Global variable to store the parsed arguments and "flag" arguments
+// Global variable to store the parsed commandline arguments
 var (
-	meshPeer = flag.String("meshPeers", "",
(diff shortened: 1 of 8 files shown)


cmd/federation-controller/main.go
```diff
@@ -5,35 +5,62 @@ import (
...
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
-	"k8s.io/klog/v2"
 )
 
-// Global variable to store the parsed arguments and "flag" arguments
+// Global variable to store the parsed commandline arguments.
 var (
-	meshPeer = flag.String("meshPeers", "",
+	meshPeer, exportedServiceSet, importedServiceSet *string
```
jewertow, line 31:
```suggestion
	meshPeer, exportedServiceSet, importedServiceSet string
```
These pointers are uninitialized, so I am getting nil pointer dereference error when I execute federation-controller:
```
make
./out/federation-controller --meshPeers '{}' --exportedServiceSet '{}'
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x1840664]

goroutine 1 [running]:
github.com/spf13/pflag.newStringValue(...)
        /home/jewertow/go/pkg/mod/github.com/spf13/pflag@v1.0.5/string.go:7
github.com/spf13/pflag.(*FlagSet).StringVar(...)
        /home/jewertow/go/pkg/mod/github.com/spf13/pflag@v1.0.5/string.go:37
main.addFlags(0xc0002c0008)
        /home/jewertow/oss/federation/cmd/federation-controller/main.go:54 +0x24
main.NewRootCommand()
        /home/jewertow/oss/federation/cmd/federation-controller/main.go:48 +0x71
main.main()
        /home/jewertow/oss/federation/cmd/federation-controller/main.go:123 +0x34
```

jewertow (changes requested):
I expected to have the ability to control particular loggers in the application, e.g. mcp, fds, fdsc, default, etc. so could you:
1) Use named loggers in particular components.
2) Add an argument to control log level of these components, like `--log_output_level`.

yxun:
@jewertow 

> I expected to have the ability to control particular loggers in the application, e.g. mcp, fds, fdsc, default, etc. so could you:

I updated this PR with logger scopes as follows:
- cmd/main.go uses `default` scope
- pkg/mcp export_hander, import_handler, resource_controller use `mcp` scope
- pkg/xds/adsc uses `adsc` scope
- pkg/xds/adss uses `adss` scope 

> 1. Use named loggers in particular components.

Do you want different variable names instead of `log` in each package/each scope ?

> 2. Add an argument to control log level of these components, like `--log_output_level`.

This has been included in cmd main.go line#63 

```
        // Attach the Istio logging options to the command.
	loggingOptions.AttachCobraFlags(c)
```
reference: https://github.com/istio/istio/blob/master/pkg/log/options.go#L201

I will add some e2e/main_tests in the next commit.

cmd/federation-controller/main.go
yxun: Fixed by following the suggested change. 
I will add some e2e main_tests to cover this case in the next commit.

jewertow (approved):
Good job, thanks!

PR #39, opened 2024-09-12
ci: introduces basic unit test job

bartoszmajsak:
In order to ensure healthy code base we can automate checks as GitHub actions. This way we can spot defects early and make sure all new changes are not breaking anything.

This change brings a simple job which for each PR or push to main branch will trigger unit tests and binary compilation (default target).

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,0 +1,12 @@
+name: controller-build
+on: [push, pull_request]
+
+jobs:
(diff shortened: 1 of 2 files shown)


bartoszmajsak:
You can test it locally using [act](https://github.com/nektos/act)

```sh
act "on: push" -j unit-test -P ubuntu-latest=ghcr.io/catthehacker/ubuntu:act-latest   
```

PR #54, opened 2024-09-30
Introduce k8s config mode

### LINKED ISSUE #50: Introduce `k8s` mode and manage Istio resources using `istio.io/istio/pkg/kube` client
**Background context:**

Federation controller provides configurations for Istio as a MCP-over-XDS server. However, this protocol has a few limitations:
1. Config source API has unimplemented TLS settings.
2. Config source must be deployed before istiod, because it will not become ready until connected to the config source (this will cause "the chicken or the egg" problem when `tls.mode: ISTIO_MUTUAL` is set).
3. Does not support delta XDS yet.

Therefore we should have an alternative mode in which the controller does not run as an MCP server, but creates, deletes, updates and reconciles Istio resources in k8s api-server. Then we will be able to enable Istio mTLS without any modification of the config source API and will not have to worry about performance of the SotW XDS.

**TODO:**
1. Add command flag, like `--mode=<"mcp" or "k8s">` or `--mode=<"xds" or "k8s">` and run the MCP server conditionally.
2. If the mode is `k8s`, then create Istio kube client to manage Istio resources:
```
import (
        "istio.io/istio/pkg/kube"
        "istio.io/client-go/pkg/apis/networking/v1alpha3"
)

func main() {
        ...
        istioClient, err := kube.NewClient(kube.NewClientConfigForRestConfig(kubeConfig), "cluster-id")
        _, err := g.client.Istio().NetworkingV1alpha3().Gateways(gateway.Namespace).Create(v1alpha3.Gateway{}, ...)
}
```

Reconciliation of the created objects can be implemented in next iterations.

eoinfennessy:
Closes #50

Current limitations:
1. Resources are not cleaned up when controller stops. We will need to add an ownerReference to resources so kube can garbage collect them.
2. Reconcile is not triggered when a user manually updates a resource managed by federation. (we will need an informer and handler similar to [this](https://github.com/openshift-service-mesh/federation/blob/master/cmd/federation-controller/main.go#L149))
3. Reconcilers GET and LIST resources by querying kube-apiserver. We should have a cached lister [similar to this](https://github.com/openshift-service-mesh/federation/blob/master/cmd/federation-controller/main.go#L149) for Istio resources as an optimisation.

### DIFF
--- a/cmd/federation-controller/main.go
+++ b/cmd/federation-controller/main.go
@@ -42,11 +42,12 @@ import (
 	"github.com/openshift-service-mesh/federation/internal/pkg/xds/adss"
 )
 
-// Global variable to store the parsed commandline arguments
 var (
-	meshPeer, exportedServiceSet, importedServiceSet string
-	loggingOptions                                   = istiolog.DefaultOptions()
-	log                                              = istiolog.RegisterScope("default", "default logging scope")
+	// Global variables to store the parsed commandline arguments
+	meshPeers, exportedServiceSet, importedServiceSet, configMode string
+
+	loggingOptions = istiolog.DefaultOptions()
+	log            = istiolog.RegisterScope("default", "default logging scope")
 )
 
 // NewRootCommand returns the root cobra command of federation-controller
@@ -67,12 +68,14 @@ func NewRootCommand() *cobra.Command {
 
 func addFlags(c *cobra.Command) {
 	// Process commandline args.
-	c.PersistentFlags().StringVar(&meshPeer, "meshPeers", "",
+	c.PersistentFlags().StringVar(&meshPeers, "meshPeers", "",
 		"Mesh peers that include address ip/hostname to remote Peer, and the ports for dataplane and discovery")
 	c.PersistentFlags().StringVar(&exportedServiceSet, "exportedServiceSet", "",
 		"ExportedServiceSet that include selectors to match the services that will be exported")
 	c.PersistentFlags().StringVar(&importedServiceSet, "importedServiceSet", "",
 		"ImportedServiceSet that include selectors to match the services that will be imported")
+	c.PersistentFlags().StringVar(&configMode, "configMode", "",
+		"Specifies how the federation controller provides mesh config to istiod. Valid options include \"mcp\" and \"k8s\".")
 
 	// Attach the Istio logging options to the command.
 	loggingOptions.AttachCobraFlags(c)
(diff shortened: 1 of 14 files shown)


internal/pkg/kube/service_entry_reconciler.go
```diff
@@ -0,0 +1,127 @@
...
+		}),
+	})
+	if err != nil {
+		return fmt.Errorf("failed to list service entries: %v", err)
+	}
+	oldServiceEntriesMap := make(map[types.NamespacedName]*v1alpha3.ServiceEntry, len(oldServiceEntries.Items))
+	for _, se := range oldServiceEntries.Items {
+		oldServiceEntriesMap[types.NamespacedName{Namespace: se.Namespace, Name: se.Name}] = se
+
+		log.Infof("******* oldSEKey: %v, oldSe: %v", oldServiceEntriesMap[types.NamespacedName{Namespace: se.Namespace, Name: se.Name}], &se.Spec)
```
jewertow, line 74: This could be logged at debug level without the wildcards.

internal/pkg/kube/reconciler_manager.go
```diff
@@ -0,0 +1,74 @@
...
+	for {
+		select {
+		case <-ctx.Done():
+			break loop
+
+		case pushRequest := <-rm.pushRequests:
+			log.Infof("[kube] Received push request: %v", pushRequest)
+
+			if r, ok := rm.reconcilers[pushRequest.TypeUrl]; !ok {
+				log.Infof("[kube] No reconciler present for type: %v", pushRequest.TypeUrl)
```
jewertow, line 65: We do not need `[kube]` prefix, because the logger is already named `kube`, and its name will be included in the log entry.

cmd/federation-controller/main.go
```diff
@@ -206,18 +217,49 @@ func main() {
...
+			kube.NewDestinationRuleReconciler(istioClient, istioConfigFactory),
+		)
+		if err := rm.ReconcileAll(ctx); err != nil {
+			log.Fatalf("initial Istio resource reconciliation failed: %v", err)
+		}
+		go rm.Start(ctx)
+
+		// TODO: implement an appropriate wait allowing proxy to receive all initial
+		// federation config before starting FDS client.
+		time.Sleep(time.Second * 5)
```
jewertow, line 255: I do not think we need to wait, because the client will try to reconnect in case of failures.

jewertow (approved):
I left a few minor comments, but I can address them in a follow-up PR.

internal/pkg/kube/service_entry_reconciler.go
eoinfennessy: Ah, sorry. I was using this to debug something and forgot to delete it.

cmd/federation-controller/main.go
eoinfennessy: As far as I remember, the connection was never established without this wait.

eoinfennessy: I think it failed [here](https://github.com/openshift-service-mesh/federation/blob/master/internal/pkg/xds/adsc/adsc.go#L70).
Maybe it would be possible to call `a.reconnect` when an error is returned here.

PR #57, opened 2024-10-16
Update chart and examples

jewertow:


examples/README.md
```diff
@@ -122,6 +93,34 @@ helm-west upgrade west-mesh chart -n istio-system \
...
+helm-east upgrade east-mesh chart -n istio-system \
+  --values examples/federation-controller.yaml \
+  --set "federation.meshPeers.remote.addresses[0]=$WEST_GATEWAY_IP" \
+  --set "federation.configMode=k8s"
+EAST_GATEWAY_IP=$(keast get svc federation-ingress-gateway -n istio-system -o jsonpath='{.status.loadBalancer.ingress[0].ip}')
+helm-west upgrade west-mesh chart -n istio-system \
+  --values examples/federation-controller.yaml \
+  --set "federation.meshPeers.remote.addresses[0]=$EAST_GATEWAY_IP" \
+  --set "federation.configMode=k8s"
+```
```
eoinfennessy, line 122: These two steps can be combined into one `helm install` step.

chart/values.yaml
```diff
@@ -3,6 +3,7 @@ image:
   tag: latest
 
 federation:
+  configMode: mcp
```
eoinfennessy, line 6: A default for this is specified in `templates/deployment.yaml`:
```
- '--configMode={{ .Values.federation.configMode | default "mcp" }}'
```

Do we want to remove the default in `templates/deployment.yaml` and have users explicitly specify it if providing their own `values.yaml`?

eoinfennessy (approved):
Looks good. I added two small comments.

chart/values.yaml
jewertow: Thanks, I forgot to remove this default value. The main reason why I added `configMode` to helm values is the conditions for annotations. If I didn't add `configMode` to values, I would have to change that condition to:
```
{{- if ne .Values.configMode "k8s" }}
```
And I thought that it will be cleaner if the default value will be defined in the default values.yaml and the condition will be simpler.

examples/README.md
jewertow: Thanks! Done.

eoinfennessy (approved):

PR #58, opened 2024-10-21
ci: restore job for unit tests

jewertow:


.github/workflows/build.yaml
```diff
@@ -0,0 +1,15 @@
+name: ci
+on:
+  push:
+    branches:
+    - master
+  pull_request:
```
eoinfennessy, line 6: Do we need to run this after pushing to the master branch? Or should it just run on PRs targeting the master branch?

```suggestion
on:
  pull_request:
    branches:
    - master
```

jewertow: > Do we need to run this after pushing to the master branch?

Good question. We usually run post-submit jobs on master, but actually we do not have to.

> Or should it just run on PRs targeting the master branch?

It's a rare case, so I did not care about it, but ok, we can restrict it to PRs target to master.

eoinfennessy (approved):

PR #60, opened 2024-10-28
test: fix flaky test case that checks if a hostname is unreachable

jewertow:
The test case "requests to c should fail" sometimes fails, because it does not receive any response, but it expects to get an error. The hostname should be unreachable, so missing responses are also correct in this case, and this behavior may be caused by a DNS response delay.

### DIFF
--- a/test/e2e/traffic_test.go
+++ b/test/e2e/traffic_test.go
@@ -69,12 +69,15 @@ func TestTraffic(t *testing.T) {
 		})
 
 		ctx.NewSubTest("requests to c should fail").Run(func(ctx framework.TestContext) {
-			a[0].CallOrFail(t, echo.CallOptions{
+			res, err := a[0].Call(echo.CallOptions{
 				Address: fmt.Sprintf("c.%s.svc.cluster.local", appNs.Name()),
 				Port:    ports.HTTP,
-				Check:   check.Status(503),
+				Check:   check.Error(),
 				Timeout: 1 * time.Second,
 			})
+			if err == nil || res.Responses.Len() != 0 {
+				t.Fatalf("the request did not fail and got the following response: %v", res)
+			}
 		})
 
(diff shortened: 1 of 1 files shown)


test/e2e/traffic_test.go
```diff
@@ -69,12 +69,15 @@ func TestTraffic(t *testing.T) {
 		})
 
 		ctx.NewSubTest("requests to c should fail").Run(func(ctx framework.TestContext) {
-			a[0].CallOrFail(t, echo.CallOptions{
+			res, err := a[0].Call(echo.CallOptions{
```
bartoszmajsak, line 72: Looking from n00b perspective - can we give `a` a meaningful name, or maybe even better, if possible, extract `a[0]` to  its own variable (thinking of `serviceA`)?

test/e2e/traffic_test.go
```diff
@@ -69,12 +69,15 @@ func TestTraffic(t *testing.T) {
...
 				Address: fmt.Sprintf("c.%s.svc.cluster.local", appNs.Name()),
 				Port:    ports.HTTP,
-				Check:   check.Status(503),
+				Check:   check.Error(),
 				Timeout: 1 * time.Second,
 			})
+			if err == nil || res.Responses.Len() != 0 {
+				t.Fatalf("the request did not fail and got the following response: %v", res)
+			}
```
bartoszmajsak, line 80: Again, n00b lenses:

- Wouldn't `Check:   check.Error()` above imply that  if we reach this line the `err` is always `nil`?  
- What if we get an empty response? Is it possible at all? Should we `Fatalf`  in such a case too?

jewertow: Thanks! I forgot about that check. I removed it.

test/e2e/traffic_test.go
bartoszmajsak: If it's a known convention there's no reason to introduce something else other than making it confusing.

bartoszmajsak (approved):
LGTM

PR #66, opened 2024-11-05
test: upload local version of federation controller to kind clusters

jewertow:
How it works now:
1. `make e2e` - build `quay.io/maistra-dev/federation-controller:test`, uploads to nodes and sets `TAG=test`.
2. `UPLOAD_TEST_IMAGE=false make e2e` - skip building image from the source code and pull the latest image in tests.
3. `UPLOAD_TEST_IMAGE=false TAG=0.1 HUB=quay.io/jewertow` - skip building image from the source code and use custom image in tests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,9 +40,13 @@ proto:
 kind-clusters:
 	bash test/scripts/kind_provisioner.sh $(ISTIO_VERSION)
 
+.PHONY: build-test-image
+build-test-image:
+	TAG=test $(MAKE) docker-build
+
 .PHONY: e2e
 TEST_SUITES ?= mcp k8s
-e2e: kind-clusters
+e2e: build-test-image kind-clusters
 	@$(foreach suite, $(TEST_SUITES), \
 		go test -tags=integ -run TestTraffic ./test/e2e/$(suite) \
 			--istio.test.hub=docker.io/istio\
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ default: build test
 export TAG ?= latest
 export HUB ?= quay.io/maistra-dev
 export ISTIO_VERSION ?= 1.23.0
-export UPLOAD_TEST_IMAGE ?= true
+export USE_LOCAL_IMAGE ?= true
 
 .PHONY: build
 build:
@@ -42,14 +42,14 @@ kind-clusters:
 	bash test/scripts/kind_provisioner.sh $(ISTIO_VERSION)
 
 .PHONY: build-test-image
-ifeq ($(UPLOAD_TEST_IMAGE),true)
+ifeq ($(USE_LOCAL_IMAGE),true)
 build-test-image:
 	TAG=test $(MAKE) docker-build
 endif
 
 .PHONY: e2e
 TEST_SUITES ?= mcp k8s
-ifeq ($(UPLOAD_TEST_IMAGE),true)
+ifeq ($(USE_LOCAL_IMAGE),true)
 	TEST_TAG := test
 else
 	TEST_TAG := $(TAG)
--- a/Makefile
+++ b/Makefile
@@ -3,9 +3,13 @@ OUT := $(shell pwd)
 .PHONY: default
 default: build test
 
-export TAG ?= latest
 export HUB ?= quay.io/maistra-dev
 export ISTIO_VERSION ?= 1.23.0
+export BUILD_TEST_IMAGE ?= true
+
+ifeq ($(BUILD_TEST_IMAGE),true)
+	export TAG := test
+endif
 
 .PHONY: build
 build:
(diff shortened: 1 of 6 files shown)


eoinfennessy (commented):
I wonder would it be cleaner to use the default `HUB` and `TAG` variables in the Makefile (or user overrides if provided) everywhere in this PR? This would involve something like the following:
- set the `HUB` and `TAG` variables when running `kind_provisioner.sh` and use the variables where `kind load` is run
- set the `HUB` and `TAG` variables when  running `go test` and continue to use them in the `helm upgrade` command
- remove `image.tag.test` from `federation_controller.yaml`
- delete `build-test-image` target and instead use `docker-build`

jewertow:
> set the `HUB` and `TAG` variables when running `kind_provisioner.sh` and use the variables where `kind load` is run

I do not see a use case to customize name for the test image. I intentionally use TAG=test, because when tag is `latest` and the `imagePullPolicy` is not set explicitly, then Kubernetes always pulls the image. But if we will change it permanently or someone will customize the pull policy to `Always`, we would not notice that we do not use the local version of the controller.

> set the `HUB` and `TAG` variables when running go test and continue to use them in the `helm upgrade` command

I restored using `HUB` and `TAG` in the test. I forgot that I intentionally used `--set` for these options.

> remove `image.tag.test` from `federation_controller.yaml`

Done.

> delete `build-test-image` target and instead use `docker-build`

I think the first reply answers this point.

eoinfennessy:
> I intentionally use TAG=test, because when tag is `latest` and the `imagePullPolicy` is not set explicitly, then Kubernetes always pulls the image. Our default is "IfNotPresent", but if we will change it in the future or someone will customize the pull policy to `Always`, we would not notice that we do not use the local version of the controller.
> 
> >

Ah, TIL images with the `latest` tag were treated like this! Thanks! My comment was more about having a single source for the `HUB` and `TAG` values that gets passed appropriately to the script/test. I can now see the benefit of hard-coding the `TAG` variable, and I suppose there's no real need to customise the `HUB` variable since this will always be a local image.

test/e2e/common/setup.go
```diff
@@ -56,7 +56,7 @@ var (
 	rootDir       = filepath.Join(filepath.Dir(file), "../../..")
 
 	testHub = env.GetString("HUB", "quay.io/maistra-dev")
-	testTag = env.GetString("TAG", "latest")
+	testTag = env.GetString("TAG", "test")
```
eoinfennessy, line 59: Now I just wonder does it make sense to now make these values constants? They are hard-coded in `kind_provisioner.sh`, so why take them from the environment here?

jewertow: > They are hard-coded in kind_provisioner.sh, so why take them from the environment here?

Because you can run `go test -tags=integ -run TestTraffic ./test/e2e/k8s ...`. However, it was incorrect and I pushed fixed version with new environment variable.

README.md
```diff
@@ -50,3 +50,7 @@ by different teams in different clusters.
     ```shell
     TEST_SUITES="k8s mcp" make e2e
     ```
+1. Customize federation controller image used in tests (`TAG` is ignored if `BUILD_TEST_IMAGE=true` or not set):
+   ```shell
+   BUILD_TEST_IMAGE=false TAG=0.1 make e2e
```
eoinfennessy, line 55: I think it would be helpful to add `HUB` to this example for completeness.

Makefile
```diff
@@ -3,9 +3,13 @@ OUT := $(shell pwd)
...
 default: build test
 
-export TAG ?= latest
 export HUB ?= quay.io/maistra-dev
 export ISTIO_VERSION ?= 1.23.0
+export BUILD_TEST_IMAGE ?= true
+
+ifeq ($(BUILD_TEST_IMAGE),true)
+	export TAG := test
+endif
```
eoinfennessy, line 12: I know that the `latest` fallback is used in the test file if not set in environment, but do you think it might be easier for future readers to do something like this?
```suggestion
ifeq ($(BUILD_TEST_IMAGE),true)
	export TAG := test
else
    export TAG ?= latest
endif
```

eoinfennessy (approved):
Looks good. I added two minor suggestions.

Makefile
jewertow: I realized that overwriting `TAG` globally depending on `BUILD_TEST_IMAGE` was a breaking change. Please look once again.

README.md
jewertow: Done

Makefile
```diff
@@ -3,13 +3,10 @@ OUT := $(shell pwd)
...
 
+export TAG ?= latest
 export HUB ?= quay.io/maistra-dev
 export ISTIO_VERSION ?= 1.23.0
-export BUILD_TEST_IMAGE ?= true
-
-ifeq ($(BUILD_TEST_IMAGE),true)
-	export TAG := test
-endif
+export UPLOAD_TEST_IMAGE ?= true
```
eoinfennessy, line 9: `UPLOAD_TEST_IMAGE` to me sounds like the image could be pushed to a registry somewhere. How about naming it `USE_LOCAL_IMAGE`, or `USE_REMOTE_IMAGE` with a default of `false`?

Feel free to merge without if you don't think this makes sense.

eoinfennessy (approved):

Makefile
jewertow: Ok, I changed `UPLOAD_TEST_IMAGE` to `USE_LOCAL_IMAGE`.

PR #68, opened 2024-11-06
fix(make): ensures test image is built for kind setup

bartoszmajsak:
Executing `make kind-clusters` on a clean environment results with the error:

```sh
+ for cluster_name in "east" "west"
+ kind load docker-image --nodes east-control-plane --name east quay.io/maistra-dev/federation-controller:test
ERROR: image: "quay.io/maistra-dev/federation-controller:test" not present locally
make: *** [Makefile:42: kind-clusters] Error 1
```

as the image tagged with `test` is not built.

This fix ensures that we build the image when `USE_LOCAL_IMAGE` is set to `true` so kind setup can succeed.


jewertow (approved):
Sorry for introducing regression. Thanks!

bartoszmajsak:
> Sorry for introducing regression. Thanks!

Evidently happens to the best of us :) No casualties though, spotted fairly quick.

PR #69, opened 2024-11-07
feat(make): introduces help target

bartoszmajsak:
As we have multiple targets defined having an easy overview of what is available without a need of skimming through `Makefile` can help newcomers.

For this purpose `make help` target has been introduced. Invoking it will print selected targets and their respective comments, giving a structured overview of available commands.

```sh
❯ make help

Usage:
  make <target>

Build
  build                      Builds the project
  test                       Runs tests
  docker-build               Builds container image
  docker-push                Pushes container image to the registry
  docker                     Combines build and push targets

Development
  kind-clusters              Provisions KinD clusters for local development or testing
  build-test-image           Builds test image

Code Gen
  proto                      Generates Go files from protobuf-based API files
  fix-imports                Fixes imports
  add-license                Adds license to all Golang files
```

Targets are grouped into sections marked by `##@ Section` and display comments added after their declarations with `##`.

> [!NOTE]
> We can also print available variables using similar pattern matching, but I am not sure how beneficial such information is if we only want a brief overview of what we can invoke.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,46 +3,47 @@ OUT := $(shell pwd)
 .PHONY: default
 default: build test
 
-export TAG ?= latest
 export HUB ?= quay.io/maistra-dev
+export TAG ?= latest
 export ISTIO_VERSION ?= 1.23.0
 export USE_LOCAL_IMAGE ?= true
 
+.PHONY: help
+help:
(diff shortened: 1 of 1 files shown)


jewertow (approved):
Nice!

eoinfennessy (approved):

PR #71, opened 2024-11-11
chore(docs): aliases for istioctl

bartoszmajsak:
Similarly to `kubectl` and `helm` we can provide aliases for `istioctl` targeting `west` and `east` clusters instead of repeating `--kubeconfig` every time we want to run some `istioctl` commands such as `install`.

This change introduces both aliases for our demo script.


sridhargaddam (approved):

bartoszmajsak:
Thanks for the review @sridhargaddam. 

I think we should rely on automation for merging moving forward, I think we follow this approach in all the repos. It seems that we have been favoring squashes for linear history and that's how the bot is set in this repo too.

We have prow/tide plugin which takes care of that, the status hints to using labels: `Not mergeable. Must be by author openshift-service-mesh-bot OR Needs okay to merge label.`

bartoszmajsak:
@jewertow I disabled merge commits in the repo settings to prevent from merge commits in the future to be consisent with bot setup. I think we missed it earlier.

PR #72, opened 2024-11-12
fds: restart client on connection failure during startup

jewertow:


cmd/federation-controller/main.go
```diff
@@ -212,6 +212,7 @@ func main() {
 			go func() {
 				if err := federationClient.Run(); err != nil {
 					log.Fatalf("failed to start FDS client: %v", err)
+					time.AfterFunc(5*time.Second, federationClient.Restart)
```
bartoszmajsak, line 215: I think we have overlooked a side-effect of calling `log.Fatalf`. It will call `os.Exit(1)` deep under the hood. 

Considering the scope of this PR, we should change the log level here to `Errorf` to restart to happen.

However, I would suggest disallowing using `log.Fatalf` due to this hidden side effect and enforcing handling errors explicitly across the code base (in this case, calling os.Exit). The latter approach also gives us an opportunity to use exit statuses other than generic `1`, for example, `128`, which indicates incorrect arguments passed to the command. I can open an issue if we agree on that.

internal/pkg/xds/adsc/adsc.go
```diff
@@ -80,6 +80,15 @@ func (a *ADSC) Run() error {
 	return nil
 }
 
+func (a *ADSC) Restart() {
+	log.Infof("reconnecting to ADS server %s", a.cfg.DiscoveryAddr)
+	err := a.Run()
+	if err != nil {
+		log.Errorf("failed to Restart to ADS server %s: %v", a.cfg.DiscoveryAddr, err)
+		time.AfterFunc(reconnectDelay, a.Restart)
+	}
```
bartoszmajsak, line 89: Nit, you can scope the error like so:

```suggestion
	if err := a.Run(); err != nil {
		log.Errorf("failed to Restart to ADS server %s: %v", a.cfg.DiscoveryAddr, err)
		time.AfterFunc(reconnectDelay, a.Restart)
	}
```

bartoszmajsak: Maybe worth adding in the log that it will attempt reconnecting after `reconnectDelay`.

cmd/federation-controller/main.go
bartoszmajsak: Normally I would give you a code `suggestion` to apply, but it affects an unedited line. Not possible on GH (yet?).

bartoszmajsak (changes requested):

cmd/federation-controller/main.go
eoinfennessy, line 215: I wonder if it would be worth adding a configurable retry count to `Restart` so that we can exit with an error instead of failing indefinitely?
```suggestion
					// Configurable retry count added to signature of Restart
					if err := federationClient.Restart(5); err != nil {
					    // Log error and exit with non-zero status 
					}
```

internal/pkg/xds/adsc/adsc.go
eoinfennessy, line 89: General idea related to comment above:
```suggestion
func (a *ADSC) Restart(retryCount int) error {
	log.Infof("reconnecting to ADS server %s with %d attempt(s) remaining", a.cfg.DiscoveryAddr, retryCount)
	err := a.Run()
	if err != nil {
		if retryCount == 1 {
            return fmt.Errorf("failed to Restart ADS client %s: %w", a.cfg.DiscoveryAddr, err)
        }
	    restartFunc := func() { a.Restart(retryCount-1) }
		time.AfterFunc(reconnectDelay, restartFunc)
	}
```

cmd/federation-controller/main.go
jewertow: > I think we have overlooked a side-effect of calling log.Fatalf. It will call os.Exit(1) deep under the hood.

I was aware of the behavior of `log.Fatalf` - I just forgot to remove it in this PR, even though that was the main goal... Initially, I decided to use `Fatalf` to exit and trigger restart of the pod.

internal/pkg/xds/adsc/adsc.go
jewertow: Done.

jewertow: `Restart` is also used when a connection is interrupted in `handleRecv`, and we should always restart the client in such a case.

cmd/federation-controller/main.go
eoinfennessy: That makes sense. Retrying indefinitely seems like a better option. Related thought: Maybe in the future we could add a status condition to the CRD indicating the connection status so that users would not need to check logs to see that the connection has failed.

internal/pkg/xds/adsc/adsc.go
```diff
@@ -80,6 +80,14 @@ func (a *ADSC) Run() error {
 	return nil
 }
 
+func (a *ADSC) Restart() {
+	log.Infof("reconnecting to ADS server %s", a.cfg.DiscoveryAddr)
+	if err := a.Run(); err != nil {
+		log.Errorf("failed to start ADS server %s, will reconnect in %s: %v", a.cfg.DiscoveryAddr, a.cfg.ReconnectDelay, err)
```
eoinfennessy, line 86: Or "failed to start ADS client"?
```suggestion
		log.Errorf("failed to connect to ADS server %s, will reconnect in %s: %v", a.cfg.DiscoveryAddr, a.cfg.ReconnectDelay, err)
```

eoinfennessy (approved):

bartoszmajsak (approved):

PR #73, opened 2024-11-13
feat(k8s): run all reconcilers in a single pass

bartoszmajsak:
Runs all reconcilers in a single pass, allowing errors to be collected without early termination. It aligns with the "eventual consistency" nature of reconciling but also enables to either set up more in a single run or capture all problems and troubleshoot more efficiently.

By using `errors.Join` we can achieve "multi-error" semantic without using any additional library such as [hashicorp/go-multierror](https://github.com/hashicorp/go-multierror).

### DIFF
--- a/internal/pkg/kube/reconciler_manager.go
+++ b/internal/pkg/kube/reconciler_manager.go
@@ -16,6 +16,7 @@ package kube
 
 import (
 	"context"
+	"errors"
 
 	istiolog "istio.io/istio/pkg/log"
 
@@ -42,12 +43,13 @@ func NewReconcilerManager(pushRequests <-chan xds.PushRequest, reconcilers ...Re
 }
 
 func (rm *ReconcilerManager) ReconcileAll(ctx context.Context) error {
+	reconcileErrs := make([]error, 0, len(rm.reconcilers))
+
 	for _, r := range rm.reconcilers {
-		if err := r.Reconcile(ctx); err != nil {
-			return err
-		}
+		reconcileErrs = append(reconcileErrs, r.Reconcile(ctx))
 	}
-	return nil
+
+	return errors.Join(reconcileErrs...)
 }
 
 func (rm *ReconcilerManager) Start(ctx context.Context) {


eoinfennessy (approved):
Great improvement! Thanks!

internal/pkg/kube/reconciler_manager.go
```diff
@@ -42,12 +43,15 @@ func NewReconcilerManager(pushRequests <-chan xds.PushRequest, reconcilers ...Re
 }
 
 func (rm *ReconcilerManager) ReconcileAll(ctx context.Context) error {
+	var reconcileErrs error
+
 	for _, r := range rm.reconcilers {
 		if err := r.Reconcile(ctx); err != nil {
-			return err
+			reconcileErrs = errors.Join(err)
```
eoinfennessy, line 50: If there are multiple errors should they all be joined?
```suggestion
			reconcileErrs = errors.Join(err, reconcileErrs)
```

internal/pkg/kube/reconciler_manager.go
```diff
@@ -42,12 +43,13 @@ func NewReconcilerManager(pushRequests <-chan xds.PushRequest, reconcilers ...Re
 }
 
 func (rm *ReconcilerManager) ReconcileAll(ctx context.Context) error {
+	var reconcileErrs []error
```
eoinfennessy, line 46: Nit: We could avoid potential reallocation by specifying a capacity.

```suggestion
	reconcileErrs := make([]error, 0, len(rm.reconcilers))
```

eoinfennessy (approved):
Looks good. One nit; feel free to ignore.

internal/pkg/kube/reconciler_manager.go
bartoszmajsak: We should probably make it `make([]error, 0, len(rm.reconcilers)))`, so the length is set correctly to `0` but capacity is reflecting how many calls we are going to make. Otherwise `append` will immediately result in doubling the capacity and so creating the new underlying array IIRC.

However, more often than not I see an nil slice used for such simple cases. I'm fine either way though. WDYT @jewertow?

jewertow: > WDYT @jewertow?

I am open to these kinds of micro optimizations. I did not pay attention to such details, because it was irrelevant in a poc, but if we can do something better, why not?

jewertow (approved):

PR #75, opened 2024-11-18
Drop support for MCP mode

jewertow:
We will no longer develop MCP (config source) server for Istio, due to the following issues:
1. No TLS support.
2. Dependency cycle in case of deploying with sidecar.
3. No delta XDS.

All of these limitations could be fixed in Istio, but we decided to develop standard k8s controller,
because it will work with Kiali out of the box and we will not need any additional interface for debugging.


internal/pkg/xds/adss/grpc_server.go
```diff
@@ -57,29 +49,28 @@ func NewServer(opts *ServerOpts, pushRequests <-chan xds.PushRequest, onNewSubsc
...
-		opts:         opts,
 	}
 }
 
 // Run starts the gRPC server and the controllers.
 func (s *Server) Run(ctx context.Context) error {
 	var routinesGroup errgroup.Group
 
-	listener, err := net.Listen("tcp", fmt.Sprintf(":%d", s.opts.Port))
+	listener, err := net.Listen("tcp", ":15080")
```
bartoszmajsak, line 59: Probably not in the scope of the PR, but IIUC we are making it configurable through

https://github.com/openshift-service-mesh/federation/blob/62123baa1e0f4e67f77803aa244f4a7e717e5e6a/internal/pkg/config/federation_config.go#L19

https://github.com/openshift-service-mesh/federation/blob/62123baa1e0f4e67f77803aa244f4a7e717e5e6a/internal/pkg/config/federation_config.go#L71-L76

Should we capture it as an issue to ensure it's propagated?

cmd/federation-controller/main.go
```diff
@@ -210,56 +202,34 @@ func main() {
...
+	if err != nil {
+		log.Fatalf("failed to create Istio client: %v", err)
+	}
 
-		if startFDSClient != nil {
-			log.Info("Starting FDS client")
-			startFDSClient()
-		}
+	rm := kube.NewReconcilerManager(
+		meshConfigPushRequests,
```
eoinfennessy, line 213: It doesn't need to be addressed here, but currently only the `TypeUrl` of the push request is used in k8s mode. The type of the `meshConfigPushRequests` channel could now just be `chan string`.

internal/pkg/xds/adss/grpc_server.go
eoinfennessy, line 59: There may not currently be a need to customise this port number, but maybe `port` could be added to the `Server` type and its constructor function?

eoinfennessy (approved):
Looks good. I added two minor comments.

cmd/federation-controller/main.go
```diff
@@ -210,56 +202,34 @@ func main() {
...
-			kube.NewServiceEntryReconciler(istioClient, istioConfigFactory),
-			kube.NewWorkloadEntryReconciler(istioClient, istioConfigFactory),
-			kube.NewVirtualServiceReconciler(istioClient, istioConfigFactory),
-			kube.NewDestinationRuleReconciler(istioClient, istioConfigFactory),
-		)
-		if err := rm.ReconcileAll(ctx); err != nil {
-			log.Fatalf("initial Istio resource reconciliation failed: %v", err)
-		}
-		go rm.Start(ctx)
+	istioClient, err := istiokube.NewClient(istiokube.NewClientConfigForRestConfig(kubeConfig), "")
```
bartoszmajsak, line 207: Little nit, but I wonder if it makes sense to pass `ISTIO_META_CLUSTER_ID` here? If so - let's open an issue.

bartoszmajsak (approved):
These are my favorite kinds of PRs!  LGTM. I also like the fact that you improved the docs! :)

I added two comments which can result in side quests (new issues).

cmd/federation-controller/main.go
jewertow: Thanks for pointing this out. I prefer to address this comment in a follow-up PR, because it will require little more refactoring in components, which takes this channel as an argument.

internal/pkg/xds/adss/grpc_server.go
jewertow: I intentionally hardcoded the container port, because users can customize Service port. But in the vast majority of cases, users will customize only the Gateway port.

cmd/federation-controller/main.go
jewertow: I [requested](https://github.com/openshift-service-mesh/federation/pull/54#discussion_r1791844922) Eoin to leave it empty, because `CLUSTER_ID` is used by the client only in debug logs to distinguish clusters it connects to. We only connect to the local cluster, so I do not see the need to set the cluster id.

internal/pkg/xds/adss/grpc_server.go
jewertow: Please look at this comment: https://github.com/openshift-service-mesh/federation/pull/75#discussion_r1848126094.

bartoszmajsak: The reason why I brought it up is b/c I see this port showing up around [ZTunnel](https://github.com/search?q=repo%3Aistio%2Fistio+15080&type=issues), though I can't find anything significant in the docs.

jewertow: I was [prototyping](https://github.com/openshift-service-mesh/federation/compare/master...jewertow:federation:examples-ambient) initial ambient support and this port did not cause any issue.

cmd/federation-controller/main.go
jewertow: `ISTIO_META_CLUSTER_ID` will only be set in the `istio-proxy` container, so this will end up in the empty string anyway. On the other hand, I do not like honoring additional environment variable in our container just to add the cluster name to a single debug message.

bartoszmajsak: If it's only about debug messaging then fair enough.

Issue #62, opened 2024-11-04
Identify source of discovery response


jewertow:
We must store imported services in a map where the key is a peer spiffe id and the value is a list of services. Currently, this is just an [array](https://github.com/openshift-service-mesh/federation/blob/master/internal/pkg/fds/imported_service_store.go#L26). The SPIFFE ID can be obtained from the XFCC header. In case of missing header, a response should be rejected.

**Update:**

We don't need XFCC to identify source cluster, because we know identities of remote peers and we know where we send requests for federated services, so the source is known when we are sending the initial DiscoveryRequest.

PR #76, opened 2024-11-19
helm: hardcode service name

jewertow:
Service name does not need to be configurable, and hardcoded name allows us to properly set SNI having only the identity, which includes service account, namespace and trust domain.

### DIFF
--- a/internal/pkg/istio/config_factory.go
+++ b/internal/pkg/istio/config_factory.go
@@ -64,8 +64,8 @@ func (cf *ConfigFactory) GetDestinationRules() *v1alpha3.DestinationRule {
 			TrafficPolicy: &istionetv1alpha3.TrafficPolicy{
 				Tls: &istionetv1alpha3.ClientTLSSettings{
 					Mode: istionetv1alpha3.ClientTLSSettings_ISTIO_MUTUAL,
-					// SNI must come from the configured identity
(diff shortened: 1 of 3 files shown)


eoinfennessy (approved):

internal/pkg/istio/config_factory.go
```diff
@@ -64,8 +64,8 @@ func (cf *ConfigFactory) GetDestinationRules() *v1alpha3.DestinationRule {
 			TrafficPolicy: &istionetv1alpha3.TrafficPolicy{
 				Tls: &istionetv1alpha3.ClientTLSSettings{
 					Mode: istionetv1alpha3.ClientTLSSettings_ISTIO_MUTUAL,
-					// SNI must come from the configured identity
-					Sni: fmt.Sprintf("federation-controller.%s.svc.cluster.local", cf.cfg.MeshPeers.Local.ControlPlane.Namespace),
+					// TODO: namespace should come from remote.identity.namespace
+					Sni: "federation-discovery-service.istio-system.svc.cluster.local",
```
bartoszmajsak, line 68: Maybe we can just open an issue instead of `TODO` in the code? Unless you have it coming shortly.

bartoszmajsak (approved):

internal/pkg/istio/config_factory.go
jewertow: This will be part of https://github.com/openshift-service-mesh/federation/issues/62.

PR #84, opened 2024-11-21
Add support for OpenShift Router

jewertow:
#### Simplified architecture

First of all, I removed the need to expose port 15080 in the ingress gateway for federation discovery service. That port existed, because I thought that we will need port with terminated TLS for discovery service (including discovering data plane gateways).

#### Ingress types

Ingress types allows us to support standard cross-cluster routing through east-west gateways, as well as OpenShift Router (haproxy). The ability to route traffic through the Router is a great advantage for OpenShift users, because they no longer need additional cloud-specific DNS setup for the east-west gateway, and they can reuse cluster DNS used by Router for mesh federation traffic.

When OpenShift Router is enabled, we must resolve peer's hostname and update ServiceEntries and WorkloadEntries in case of an IP change. This is why federation controller disables IP auto-allocation when this ingress type is enabled.

Currently, all peers must use the same ingress type, but this should be very easy to fix - cluster role needs additional condition for `destinationrules`, but I didn't have enough time to test this case. I will fix it in a follow-up PR.

Exposing exported service on OpenShift Route:
```
kind: Route
apiVersion: route.openshift.io/v1
metadata:
  name: <exported_svc>-<ns>-<port>-to-federation-ingress-gateway
  namespace: istio-system
spec:
  host: <exported_svc>-<port>.<ns>.svc.cluster.local
  to:
    kind: Service
    name: <gateway_port_name>
    weight: 100
  port:
    targetPort: tls-passthrough
  tls:
    termination: passthrough
  wildcardPolicy: None
---
apiVersion: networking.istio.io/v1alpha3
kind: EnvoyFilter
metadata:
  name: add-sni-to-listener
  namespace: istio-system
spec:
  workloadSelector:
    labels:
      app: federation-ingress-gateway
  configPatches:
  - applyTo: FILTER_CHAIN
    match:
      listener:
        name: "0.0.0.0_<gateway_port>"
        filterChain:
          sni: "outbound_.<exported_service_port>_._.<exported_service_fqdn>"
    patch:
      operation: MERGE
      value:
        filter_chain_match:
          server_names:
          - "<exported_service>-<port>.<ns>.svc.cluster.local"
```
Customizing SNI when openshfit-router is enabled:
```
apiVersion: networking.istio.io/v1
kind: DestinationRule
metadata:
  name: mtls-sni-federation-discovery-service-west-istio-system
  namespace: istio-system
spec:
  host: <remote_peer_hostname>
  trafficPolicy:
    tls:
      mode: ISTIO_MUTUAL
      sni: <imported_svc>-<port>.<ns>.svc.cluster.local
```

#### Peer name and PeerAuthentication

The controller was creating host alias for remote peer and was originating mTLS explicitly using a DR. Now, the FDS client connects to remote peer using its original service name, but this requires to specify the remote peer's name in the configuration to use proper hostname in FDS client. The DR is no longer need, because we create `PeerAuthentication` for the controller with STRICT mTLS.

#### What does this PR not include...

This PR does not include unit tests, but I was testing all these changes using our examples both on KinD and OpenShift clusters. I will implement unit tests in a follow-up PR.

I also figured out that importing a Service that exists locally will not work if the remote peer address is a hostname, because Istio does not support hostname in WorkloadEntries, which are not associated with ServiceEntries with DNS resolution type.

I did not cover edge cases, like too long hostname in Route - it rejects hosts longer than 63 characters. Will be addressed in another PR.


internal/pkg/istio/config_factory.go
```diff
@@ -199,74 +286,74 @@ func (cf *ConfigFactory) GetWorkloadEntries() ([]*v1alpha3.WorkloadEntry, error)
...
+				Protocol: "TLS",
+			}},
+			Location:   istionetv1alpha3.ServiceEntry_MESH_INTERNAL,
+			Resolution: istionetv1alpha3.ServiceEntry_DNS,
+		}
+	} else {
+		se.Spec = istionetv1alpha3.ServiceEntry{
+			// TODO: this will not work for ingressType=istio when the remote address is a hostname
+			Hosts:     []string{fmt.Sprintf("federation-discovery-service-%s.istio-system.svc.cluster.local", cf.cfg.MeshPeers.Remote.Name)},
+			Addresses: common.Resolve(cf.cfg.MeshPeers.Remote.Addresses[0]),
```
bartoszmajsak, line 313: Would this make sense?
```suggestion
			Addresses: slices.Flatten(slices.Map(cf.cfg.MeshPeers.Remote.Addresses, func(addr string) []string {
					return common.Resolve(addr)
			})),
```

internal/pkg/kube/gateway_reconciler.go
```diff
@@ -46,7 +46,7 @@ func (r *GatewayResourceReconciler) GetTypeUrl() string {
 }
 
 func (r *GatewayResourceReconciler) Reconcile(ctx context.Context) error {
-	gw, err := r.cf.GetIngressGateway()
+	gw, err := r.cf.IngressGateway()
```
bartoszmajsak, line 49: nit: we have `GetWorkloadEntries` left behind with the getter naming convention

internal/pkg/istio/config_factory.go
```diff
@@ -199,74 +286,74 @@ func (cf *ConfigFactory) GetWorkloadEntries() ([]*v1alpha3.WorkloadEntry, error)
...
 				Number:   15080,
 				Protocol: "GRPC",
 			}},
-			Location:   istionetv1alpha3.ServiceEntry_MESH_EXTERNAL,
+			Endpoints: slices.Map(cf.cfg.MeshPeers.Remote.Addresses, func(addr string) *istionetv1alpha3.WorkloadEntry {
+				we := &istionetv1alpha3.WorkloadEntry{
+					Address: addr,
+					Labels:  map[string]string{"security.istio.io/tlsMode": "istio"},
+					Ports:   map[string]uint32{"grpc": cf.cfg.MeshPeers.Remote.GetPort()},
+					Network: cf.cfg.MeshPeers.Remote.Network,
```
bartoszmajsak, line 324: I am not sure how these values are related or if there are significant at all, but across the entire config (e.g. Istio or Sail Operator config, `topology.istio.io/network` or `ISTIO_META_REQUESTED_NETWORK_VIEW` in Gateway config) we use `east|west-network`. Here however it comes from sample setup and we set it as `east` or `west` there.

bartoszmajsak (changes requested):
I have a few minor suggestions and couple of questions that will help me better understand this change.

I tested it locally on my side and it seems to work with my toy apps.

internal/pkg/istio/config_factory.go
jewertow: In general more than one address does not make any sense, because all of them will be resolved to the same set of IPs.

internal/pkg/istio/config_factory.go
jewertow: I noticed this inconsistency and I was surprised that I did not face any routing issue caused by this inconsistency, but of course I forgot to fix it. Done.

internal/pkg/kube/gateway_reconciler.go
jewertow: Done

internal/pkg/kube/envoy_filter_reconciler.go
```diff
@@ -0,0 +1,126 @@
...
+
+	kind := "EnvoyFilter"
+	apiVersion := "networking.istio.io/v1alpha3"
+	for k, ef := range envoyFiltersMap {
+		oldEF, ok := oldEnvoyFiltersMap[k]
+		if !ok || !reflect.DeepEqual(&oldEF.Spec, &ef.Spec) {
+			// Envoy filter does not currently exist or requires update
+			newEF, err := r.client.Istio().NetworkingV1alpha3().EnvoyFilters(ef.GetNamespace()).Apply(ctx,
+				// TODO v1alpha4?
+				&v1alpha4.EnvoyFilterApplyConfiguration{
```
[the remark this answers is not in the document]
jewertow: I added `// TODO v1alpha4?`, because I thought that I imported `v1alpha4` and I was surprised that it works, but then I realized that this is just Goland... I renamed this import.

internal/pkg/kube/destination_rule_reconciler.go
```diff
@@ -46,36 +53,73 @@ func (r *DestinationRuleReconciler) GetTypeUrl() string {
...
+				return fmt.Errorf("failed to apply destination rule: %w", err)
+			}
+			log.Infof("Applied destination rule: %v", newDR)
+		}
+	}
+
+	for k, oldDR := range oldDestinationRulesMap {
+		if _, ok := destinationRulesMap[k]; !ok {
+			err := r.client.Istio().NetworkingV1alpha3().DestinationRules(oldDR.GetNamespace()).Delete(ctx, oldDR.GetName(), metav1.DeleteOptions{})
+			if err != nil && !errors.IsNotFound(err) {
```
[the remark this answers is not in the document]
jewertow: Done

internal/pkg/common/resolve.go
```diff
@@ -0,0 +1,22 @@
+package common
+
+import (
+	"fmt"
+	"net"
+
+	"istio.io/istio/pkg/slices"
+)
+
+func Resolve(addr string) []string {
```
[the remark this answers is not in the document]
jewertow: Moved to `pkg/networking`, but I do not agree with the name `ResolveIPs`, because we resolve hostnames, not IPs, so I left `Resolve`. Is it ok?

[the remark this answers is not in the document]
bartoszmajsak: Right, we resolve addr (hostname or IP) to actual IPs, so simple `Resolve` is enough here.

internal/pkg/istio/config_factory.go
bartoszmajsak: Yeah this is why I started with a doubt if it's significant at all. Not sure if it's a bug somewhere or we have to dig deeper to see why it was not causing issues. Maybe when writing tests for next PR you can introduce inconsistency and see if there's an issue.

internal/pkg/xds/types.go
```diff
@@ -15,10 +15,12 @@
...
-	GatewayTypeUrl         = "networking.istio.io/v1alpha3/Gateway"
-	ServiceEntryTypeUrl    = "networking.istio.io/v1alpha3/ServiceEntry"
-	WorkloadEntryTypeUrl   = "networking.istio.io/v1alpha3/WorkloadEntry"
-	VirtualServiceTypeUrl  = "networking.istio.io/v1alpha3/VirtualService"
+	ExportedServiceTypeUrl    = "federation.istio-ecosystem.io/v1alpha1/ExportedService"
+	DestinationRuleTypeUrl    = "networking.istio.io/v1alpha3/DestinationRule"
+	GatewayTypeUrl            = "networking.istio.io/v1alpha3/Gateway"
+	ServiceEntryTypeUrl       = "networking.istio.io/v1alpha3/ServiceEntry"
+	WorkloadEntryTypeUrl      = "networking.istio.io/v1alpha3/WorkloadEntry"
+	EnvoyFilterTypeUrl        = "networking.istio.io/v1alpha3/EnvoyFilter"
```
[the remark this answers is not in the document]
bartoszmajsak: Created a follow-up issue #89

bartoszmajsak (approved):
LGTM, thanks for taking into account my suggestions.

internal/pkg/kube/destination_rule_reconciler.go
```diff
@@ -46,36 +50,73 @@ func (r *DestinationRuleReconciler) GetTypeUrl() string {
...
+		}),
 	})
 	if err != nil {
-		return fmt.Errorf("failed to apply destination rule: %w", err)
+		return fmt.Errorf("failed to list destination rules: %w", err)
+	}
+	oldDestinationRulesMap := make(map[types.NamespacedName]*v1alpha3.DestinationRule, len(oldDestinationRules.Items))
+	for _, se := range oldDestinationRules.Items {
+		oldDestinationRulesMap[types.NamespacedName{Namespace: se.Namespace, Name: se.Name}] = se
+	}
```
eoinfennessy, line 74:
```suggestion
	for _, dr := range oldDestinationRules.Items {
		oldDestinationRulesMap[types.NamespacedName{Namespace: dr.Namespace, Name: dr.Name}] = dr
	}
```

internal/pkg/kube/peer_authentication_reconciler.go
```diff
@@ -0,0 +1,96 @@
...
+			Selector: &typev1beta1.WorkloadSelector{
+				MatchLabels: map[string]string{
+					"app.kubernetes.io/name": "federation-controller",
+				},
+			},
+			Mtls: &securityv1beta1.PeerAuthentication_MutualTLS{
+				Mode: securityv1beta1.PeerAuthentication_MutualTLS_STRICT,
+			},
+		},
+	}
```
eoinfennessy, line 67: Is it worth adding this to the Istio `ConfigFactory` for consistency?

eoinfennessy (approved):
Looks good! I added some minor suggestions and clarifying questions.

chart/values.yaml
```diff
@@ -10,16 +10,47 @@ istio:
...
+        # Gateway for exported services, etc.).
         namespace: istio-system
+      gateways:
+        ingress:
+          # Ingress gateway selector specifies to which workloads Gateway configurations will be applied.
+          # selector:
+          #   app: federation-ingress-gateway
+          port:
+            # Port name of the ingress gateway Service.
+            # This is relevant only when the ingressType is openshift-router, but it cannot be empty.
```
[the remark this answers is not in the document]
jewertow: `port.name` is used in the `Gateway` resource managed by the controller, and if `openshift-router` is enabled, this port name is also used in `Route.spec.port.targetPort` - then the route target port must match the gateway's service port name. If the ingress type is Istio, only the port number is important and the name could be whatever you want, because it's not used by Istio.

cmd/federation-controller/main.go
```diff
@@ -189,9 +195,13 @@ func main() {
 
 	var fdsClient *adsc.ADSC
 	if len(cfg.MeshPeers.Remote.Addresses) > 0 {
+		discoveryAddr := fmt.Sprintf("federation-discovery-service-%s.%s.svc.cluster.local:15080", cfg.MeshPeers.Remote.Name, cfg.MeshPeers.Local.ControlPlane.Namespace)
+		if cfg.MeshPeers.Remote.IngressType == config.OpenShiftRouter {
+			discoveryAddr = cfg.MeshPeers.Remote.Addresses[0]
+		}
```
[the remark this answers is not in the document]
jewertow: Right, this looks better. Done

cmd/federation-controller/main.go
```diff
@@ -205,19 +215,41 @@ func main() {
...
+			log.Fatalf("failed to create Route client: %v", err)
+		}
+
+		reconcilers = append(reconcilers, kube.NewDestinationRuleReconciler(istioClient, istioConfigFactory))
+		reconcilers = append(reconcilers, kube.NewEnvoyFilterReconciler(istioClient, istioConfigFactory))
+		reconcilers = append(reconcilers, kube.NewRouteReconciler(routeClient, openshift.NewConfigFactory(*cfg, serviceLister)))
+
+		go func() {
+			log.Debugf("Resolving %s", cfg.MeshPeers.Remote.Addresses[0])
+			lastIPs := networking.Resolve(cfg.MeshPeers.Remote.Addresses[0])
```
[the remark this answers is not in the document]
jewertow: I know this looks better and that was my first approach, but this results in the log message: "IP addresses of %s have changed". And I thought it's better to not print it on the initial resolution, because if I as a user will want to check if the DNS record changes over time and triggers reconciliation I will see the message even though the IPs were not really changed.

cmd/federation-controller/main.go
```diff
@@ -205,19 +215,41 @@ func main() {
...
+				log.Debugf("Resolving %s", cfg.MeshPeers.Remote.Addresses[0])
+				ips := networking.Resolve(cfg.MeshPeers.Remote.Addresses[0])
+				sort.Strings(ips)
+				if !slices.Equal(lastIPs, ips) {
+					log.Infof("IP addresses of %s have changed", cfg.MeshPeers.Remote.Addresses[0])
+					lastIPs = ips
+					meshConfigPushRequests <- xds.PushRequest{TypeUrl: xds.ServiceEntryTypeUrl}
+					meshConfigPushRequests <- xds.PushRequest{TypeUrl: xds.WorkloadEntryTypeUrl}
+				}
+				time.Sleep(1 * time.Second)
```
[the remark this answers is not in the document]
jewertow: I think we will refactor this goroutine when we will have kube-builder and all its components, so I want simplicity for now. I also think that fine grained settings like this one should be exposed not earlier than in tech-preview release, because we may remove it or completely change in the meantime.

internal/pkg/kube/destination_rule_reconciler.go
jewertow: Copy-paste... Thanks :)

chart/values.yaml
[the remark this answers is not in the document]
eoinfennessy: Makes sense. Thanks for the clarification.

cmd/federation-controller/main.go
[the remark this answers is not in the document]
eoinfennessy: Makes sense. Thanks.

internal/pkg/kube/peer_authentication_reconciler.go
jewertow: Config factory exists only because we had MCP and K8S mode and the factory encapsulated common logic, but now it's no longer needed. @bartoszmajsak also noted that this is weird pattern. I guess we will move all of that logic to reconcilers or we will refactor that factory to functions which take required arguments and it want be part of a struct, so I didn't add anything new to that factory. WDYT?

eoinfennessy: I quite like that there is a separation of concerns between defining/generating the configuration and applying the configuration. It's a bit strange at the moment because these configs are then converted to `applyConfigurations` (largely to reduce additional calls to kube-apiserver), but this will no longer be required when we move to controller-runtime.

eoinfennessy: > I guess we will move all of that logic to reconcilers
This also seems like a good option, but I'm of the opinion we should remain consistent with the current pattern until when this refactor takes place (assuming it does).

internal/pkg/kube/destination_rule_reconciler.go
bartoszmajsak: Oh yeah, I was looking for this one as I noticed it after first read a few days ago :D But I failed after a quick scan so I assumed it was fixed in the meantime :) Thanks @eoinfennessy

internal/pkg/kube/peer_authentication_reconciler.go
bartoszmajsak: > @bartoszmajsak also noted that this is weird pattern.

I was only curious about the struct itself and naming as I would probably just go with simple funcs. I think we will keep generators in some sort of fashion, but I believe we can review/revamp it as we go.

jewertow: @eoinfennessy I will move it out of the reconciler in a PR with unit tests.

PR #90, opened 2024-11-28
feat(ci): ensures license headers are added

bartoszmajsak:
`make add-license` target is not part of the default Makefile target, therefore it can be overlooked. But even if it were, there's no guarantee it will be run before a PR is opened or updated. 

The `linters/check-license` GitHub job verifies that every relevant file includes
the required license headers. Otherwise, job fails and lists files both in the [logs](https://github.com/bartoszmajsak/ossm-federation/actions/runs/12066829341/job/33648514031?pr=1)
of the job as well as in the [summary](https://github.com/bartoszmajsak/ossm-federation/actions/runs/12066829341?pr=1#summary-33648514031).

### DIFF
--- a/.github/workflows/linters.yaml
+++ b/.github/workflows/linters.yaml
@@ -1,0 +1,25 @@
+name: linters
+on:
+  pull_request:
+    branches:
+    - master
+jobs:
+  check-license:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+    - name: Check licenses
+      run: | 
+        make add-license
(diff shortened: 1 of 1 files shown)


bartoszmajsak:
Tested on my fork (links in the description)

eoinfennessy (approved):
Looks good! Can you please also rename `.github/workflows/build.yaml` in this PR?

bartoszmajsak:
> Looks good! Can you please also rename `.github/workflows/build.yaml` in this PR?

I think it goes beyond the scope of this PR. We can raise another one with this tiny change.

jewertow (approved):

PR #91, opened 2024-11-28
feat(chart): add "app" label to enable app recognition in Kiali

jewertow:
Kiali uses `app` label to associate services and workloads as applications.

**Before:**
kiali-before

**After:**
kiali-after

### DIFF
--- a/chart/templates/deployment.yaml
+++ b/chart/templates/deployment.yaml
@@ -22,7 +22,6 @@ spec:
       {{- end }}
       labels:
         {{- include "chart.labels" . | nindent 8 }}
-        app.kubernetes.io/name: federation-controller
         sidecar.istio.io/inject: "true"
     spec:
       serviceAccountName: {{ include "chart.name" . }}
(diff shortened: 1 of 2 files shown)


chart/templates/deployment.yaml
```diff
@@ -22,7 +22,6 @@ spec:
       {{- end }}
       labels:
         {{- include "chart.labels" . | nindent 8 }}
-        app.kubernetes.io/name: federation-controller
```
bartoszmajsak, line 25: Is it intentional removal? I would keep standard k8s labels.

bartoszmajsak:
@jewertow I find it a bit odd that it does not work with the standard label `app.kubernetes.io/name`. Do you know why `app` is preferred?

chart/templates/deployment.yaml
jewertow: I removed it from this template, because it's already included in `chart.selectorLabels`, which are included in `chart.labels`.

jewertow:
I found a few related issues and discussions in Kiali, but I still do not know how can I configure Kiali to make it work.
Please do not merge it yet. I opened this PR too early. I will investigate it more or will ask them directly.

bartoszmajsak:
@jewertow maybe we should introduce so sort of `on-hold` label? This would prevent (hopefully) from accidental manual merges and we can configure `tide` to not auto-merge where such a label exists. Alternatively you could convert this PR back to draft to avoid confusion.

PR #92, opened 2024-11-28
feat(build): simplifies local image upload strategy

bartoszmajsak:
When we want to use a locally built image for testing, instead of relying on the arbitrary `test` tag, we can follow a convention of appending `-local` suffix to any tag configured through the `TAG` variable. Such an appended tag is only used when provisioning KinD clusters and running e2e tests.

This PR simplifies the process of uploading the image to KinD cluster by moving the decision of performing it outside of the function itself. The function now accepts `cluster`, `hub` and `tag` as *ordered* parameters. The last two are defaulted when not passed or passed as empty (`""`).

The scripts have been cleaned up and some functions moved to their own files.

Any function in `lib.sh` and `cluster.sh` can now be invoked directly from the shell. So you can e.g. update image in the running cluster.

<details>
<summary>Using upload_image from shell</summary>

```shell
❯ ./test/scripts/lib.sh upload_image west
Image with ID: sha256:ed8b3ed32e2d14ebafef7f63fdabfb648cefcc5a986df8f24f76f44361cf6a0c already present on the node west-control-plane but is missing the tag quay.io/maistra-dev/federation-controller:latest. re-tagging...

❯ ./test/scripts/lib.sh upload_image west quay.io/bartoszmajsak test-tag
Image: "quay.io/bartoszmajsak/federation-controller:test-tag" with ID "sha256:83595794b2d74dabbfc975a07bf7f6ced0d9e3d9e3aea50a9f1e6704c7fa2478" not yet present on node "west-control-plane", loading...

❯ ./test/scripts/lib.sh upload_image west "" test-tag
Image with ID: sha256:83595794b2d74dabbfc975a07bf7f6ced0d9e3d9e3aea50a9f1e6704c7fa2478 already present on the node west-control-plane but is missing the tag quay.io/maistra-dev/federation-controller:test-tag. re-tagging...

```

> [!NOTE]
> The function relies on `HUB` and `TAG` env variables values if parameters
> are not passed explicitly. If these are not specified, default values
are used. But if you want to just override `TAG`, pass "" for the  second parameter
> (`HUB`) like in the last example above.

</details>

### DIFF
--- a/test/scripts/cluster.sh
+++ b/test/scripts/cluster.sh
@@ -1,0 +1,84 @@
+#!/bin/bash
+
+ROOT=$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)
+
+function create_kind_cluster {
+  cluster=$1
+  podSubnet=$2
+  svcSubnet=$3
+
+  echo "Creating cluster '$cluster' in podSubnet='$podSubnet' and svcSubnet='$svcSubnet'"
+
+  kind create cluster --name "$cluster" --config=<<EOF
+apiVersion: kind.x-k8s.io/v1alpha4
+kind: Cluster
+networking:
+  podSubnet: "$podSubnet"
+  serviceSubnet: "$svcSubnet"
+EOF
+}
+
+function provision_kind_clusters {
+  echo "Creating KinD clusters"
+  kind_pids=()
+  create_kind_cluster east 10.10.0.0/16 10.255.10.0/24 &
+  kind_pids[0]=$!
+  create_kind_cluster west 10.30.0.0/16 10.255.30.0/24 &
+  kind_pids[1]=$!
+
+  for pid in ${kind_pids[*]}; do
+    wait $pid
+  done
+
+  kind get kubeconfig --name east > $ROOT/east.kubeconfig
+  kind get kubeconfig --name west > $ROOT/west.kubeconfig
+
+  echo "Installing MetalLB"
+  metallb_pids=()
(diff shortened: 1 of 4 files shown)


test/scripts/cluster.sh
```diff
@@ -0,0 +1,84 @@
...
+  serviceSubnet: "$svcSubnet"
+EOF
+}
+
+function provision_kind_clusters {
+  echo "Creating KinD clusters"
+  kind_pids=()
+  create_kind_cluster east 10.10.0.0/16 10.255.10.0/24 &
+  kind_pids[0]=$!
+  create_kind_cluster west 10.30.0.0/16 10.255.30.0/24 &
```
jewertow, line 26: I was checking your PR locally and I noticed that these subnets are ignored, i.e. clusters use default subnets - 10.244.0.0/16 for pods and 10.96.0.0/16 for services, but they are also ignored on master, so it's not your fault. Instead of fixing it, we can remove these subnets, because mesh federation doesn't require non-overlapping CIDRs, so these custom subnets were always irrelevant.

jewertow (changes requested):

test/scripts/cluster.sh
bartoszmajsak: @jewertow I have a follow-up PR that further expands cluster-related functions for multiple peer setups.  I think this fix is a perfect addition to that one. WDYT?

jewertow (approved):

PR #93, opened 2024-11-29
feat(networking): improve handling remote host address

jewertow:
This PR includes the following changes:

1. ServiceEntries for remote FDS are unified, i.e. we will no longer create different hosts and ports depending on the ingress type.
2. ServiceEntries use STATIC resolution if the remote addresses are IPs and DNS resolution if the address is a DNS name. This reduces the impact of IP changes on the managed resources, because we will only need to update WorkloadEntries, and Envoy will care of endpoints associated with ServiceEntries.
3. FDS client sends remote hostname in the authority pseudo-header - this allows us to remove disabling IP auto-allocation and unify `hosts` field in SE for remote FDS.
4. Permissions for DestinationRule and EnvoyFilter depend on remote and local ingress types.
5. Added unit tests for export-related Istio configs (more tests in next PRs).

The above changes add support for remote host address when remote ingress type is "istio", and decouple handling ingress type "openshift-router" with remote hostnames.


test/e2e/common/setup.go
```diff
@@ -333,3 +336,113 @@ func SetEnvAndKubeConfigPath(cmd *exec.Cmd, clusterIndex int) {
 	cmd.Env = os.Environ()
 	cmd.Env = append(cmd.Env, fmt.Sprintf("KUBECONFIG=%s/test/%s.kubeconfig", RootDir, ClusterNames[clusterIndex]))
 }
+
+func PatchCoredns(ctx resource.Context) error {
```
bartoszmajsak, line 340: How about moving it to it's own file `coredns.go`? 

Also.... as we don't have a linter yet... `PatchCoreDNS` :)

bartoszmajsak: Some small godoc as to why we are patching could help long term.

cmd/federation-controller/main.go
```diff
@@ -35,13 +43,6 @@ import (
 	"github.com/openshift-service-mesh/federation/internal/pkg/xds"
 	"github.com/openshift-service-mesh/federation/internal/pkg/xds/adsc"
 	"github.com/openshift-service-mesh/federation/internal/pkg/xds/adss"
-	routev1client "github.com/openshift/client-go/route/clientset/versioned"
-	istiokube "istio.io/istio/pkg/kube"
-	istiolog "istio.io/istio/pkg/log"
-	"istio.io/istio/pkg/slices"
-	corev1 "k8s.io/api/core/v1"
-	"k8s.io/client-go/informers"
-	"k8s.io/client-go/rest"
```
jewertow, line 44: @bartoszmajsak this is `fix-imports`, not me.

cmd/federation-controller/main.go
```diff
@@ -134,21 +135,20 @@ func main() {
 	}()
 
 	var fdsClient *adsc.ADSC
-
 	remote := cfg.MeshPeers.Remote
-
 	if len(remote.Addresses) > 0 {
-		remoteName := remote.Name
```
jewertow, line 141: @bartoszmajsak if it comes to these little formatting changes:
1. I don't see value in multiplying variables if the new ones are not notably shorter.
2. I like compact pieces of code within a function, like this:
```
// I do something
var a
var b
condition that uses above variables

// I do something else
var c
var d
func1(c)
func2(d)
```

chart/templates/clusterrole.yaml
```diff
@@ -12,9 +12,14 @@ rules:
 - apiGroups: ["security.istio.io"]
   resources: ["peerauthentications"]
   verbs: ["get", "list", "create", "update", "patch", "delete"]
+{{- if eq .Values.federation.meshPeers.remote.ingressType "openshift-router" }}
+- apiGroups: ["networking.istio.io"]
+  resources: ["destinationrules"]
+  verbs: ["get", "list", "create", "update", "patch", "delete"]
+{{- end }}
```
bartoszmajsak, line 19: Nit: could be PR on it's own as a fix or chore.

cmd/federation-controller/main.go
```diff
@@ -168,16 +168,26 @@ func main() {
...
 		reconcilers = append(reconcilers, kube.NewRouteReconciler(routeClient, openshift.NewConfigFactory(*cfg, serviceLister)))
+	}
 
+	rm := kube.NewReconcilerManager(meshConfigPushRequests, reconcilers...)
+	if err := rm.ReconcileAll(ctx); err != nil {
+		log.Fatalf("initial Istio resource reconciliation failed: %v", err)
+	}
+	go rm.Start(ctx)
+
+	if !networking.IsIP(remote.Addresses[0]) {
```
jewertow, line 190: In the future we could start the resolver only if a locally existing service was imported (will create an issue).

internal/pkg/istio/testdata/fds-envoy-filter.yaml
jewertow, line 1: `kind` and `apiVersion` are set in the reconciler. I will change that in another PR.

internal/pkg/istio/config_factory.go
```diff
@@ -160,8 +161,11 @@ func (cf *ConfigFactory) IngressGateway() (*v1alpha3.Gateway, error) {
 	return gateway, nil
 }
 
+// EnvoyFilters returns patches for SNI filters matching SNIs of exported services in federation ingress gateway.
+// These patches add SNI compatible with https://datatracker.ietf.org/doc/html/rfc952 required by OpenShift Router.
+// This function returns empty list when the local ingress type is "istio".
```
bartoszmajsak, line 166: Nit: it doesn't return empty list (or slice), but `nil` pointer. Either change godoc or implementation :)

internal/pkg/istio/config_factory_test.go
```diff
@@ -0,0 +1,306 @@
...
+						"export": "true",
+					},
+				}},
+			}},
+		},
+	}
+
+	unexportService = &corev1.Service{
+		ObjectMeta: v1.ObjectMeta{Name: "a", Namespace: "ns1"},
+	}
```
bartoszmajsak, line 73: nit: `unexported`?

internal/pkg/istio/config_factory_test.go
```diff
@@ -0,0 +1,306 @@
...
+		})
+	}
+}
+
+func TestEnvoyFilters(t *testing.T) {
+	testCases := []struct {
+		name                     string
+		localIngressType         config.IngressType
+		existingServices         []*corev1.Service
+		expectedEnvoyFilterFiles []string
```
jewertow, line 204: In case of EnvoyFilters it's better to compare results to the serialized objects, because creating proper patch values require to use tricky protobuf serialization, and we would have to copy function from the config factory, and such a test would not make any value to me.

internal/pkg/istio/config_factory_test.go
```diff
@@ -0,0 +1,306 @@
...
+	}
+)
+
+func TestIngressGateway(t *testing.T) {
+	testCases := []struct {
+		name             string
+		existingServices []*corev1.Service
+		expectedGateway  *v1alpha3.Gateway
+	}{{
+		name:             "found 2 services matching configured label selector",
```
bartoszmajsak, line 112: nit: the name is a bit confusing for me. it's more about defining 2 exported services as hosts for gateway than finding based on the selector?

cmd/federation-controller/main.go
```diff
@@ -134,21 +135,20 @@ func main() {
...
-
 	remote := cfg.MeshPeers.Remote
-
 	if len(remote.Addresses) > 0 {
-		remoteName := remote.Name
 		var discoveryAddr string
-		if remote.IngressType == config.OpenShiftRouter {
-			discoveryAddr = remote.Addresses[0]
+		if networking.IsIP(remote.Addresses[0]) {
+			discoveryAddr = fmt.Sprintf("federation-discovery-service-%s.istio-system.svc.cluster.local:15080", remote.Name)
```
bartoszmajsak, line 142: I wonder if we shouldn't encapsulate it in remote struct. The naming is something to polish further, but I was thinking about:

```golang
hostname, port := remote.Service()
discoveryAddr := fmt.Sprintf("federation-discovery-service-%s.istio-system.svc.cluster.local:%d", hostname, port)
```
where hostname can be also used as authority.

And impl:

```golang
func (r *Remote) Service() (string, uint32) {
	hostname := r.Addresses[0]
	if networking.IsIP(hostname) {
		hostname = fmt.Sprintf("federation-discovery-service-%s.istio-system.svc.cluster.local", r.Name)
	}

	return hostname, 15080
}
```

This would also limit duplication in all `config_factory.go`

internal/pkg/networking/resolve.go
```diff
@@ -21,16 +21,26 @@ import (
...
+			allIPs = append(allIPs, addr)
+			continue
+		}
 
-	ips, err := net.LookupIP(addr)
-	if err != nil {
-		fmt.Printf("Failed to resolve '%s': %v\n", addr, err)
+		ips, err := net.LookupIP(addr)
+		if err != nil {
+			fmt.Printf("Failed to resolve '%s': %v\n", addr, err)
```
bartoszmajsak, line 34: I think we shouldn't mix logging and printf. Not for this PR per se, but something to unify moving forward.

test/e2e/common/coredns/patch.go
```diff
@@ -0,0 +1,149 @@
...
+func updateCorefile(coredns *corev1.ConfigMap, hosts map[string]string, clusterName string) error {
+	const hostsTemplate = `
+    hosts {
+%s
+      fallthrough
+    }`
+
+	var hostsEntries []string
+	for name, ip := range hosts {
+		hostsEntries = append(hostsEntries, fmt.Sprintf("      %s %s", ip, name))
```
bartoszmajsak, line 125: seeing `      ` boosts my anxiety :) but for this given case it's good enough :D

test/e2e/common/setup.go
```diff
@@ -66,6 +66,11 @@ const (
 	WestClusterName = "cluster-1"
 )
 
+type InstallOptions struct {
+	EnableSpire      bool
+	SetRemoteDNSName bool
+}
```
bartoszmajsak, line 72: I have something similar in my coming PR, so will adjust when working on massive rebase xD

test/e2e/remote_dns_name/main_test.go
bartoszmajsak, line 1: Is `main_test.go` a convention in Istio community? I would think we could name it `suite_test.go` as it's test suite after all.

test/e2e/remote_dns_name/main_test.go
```diff
@@ -0,0 +1,56 @@
...
+		Setup(common.InstallOrUpgradeFederationControllers(common.InstallOptions{SetRemoteDNSName: true})).
+		Setup(namespace.Setup(&common.AppNs, namespace.Config{Prefix: "app", Inject: true})).
+		// a - client
+		// b - service available in east and west clusters - covers importing with WorkloadEntry
+		// c - service available only in west cluster - covers importing with ServiceEntry
+		Setup(common.DeployApps(&common.EastApps, common.EastClusterName, namespace.Future(&common.AppNs), false, "a", "b")).
+		Setup(common.DeployApps(&common.WestApps, common.WestClusterName, namespace.Future(&common.AppNs), false, "b", "c")).
+		// c must be removed from the east cluster, because we want to test importing a service
+		// that exists only in the remote cluster.
+		Setup(common.RemoveServiceFromClusters("c", namespace.Future(&common.AppNs), common.EastClusterName)).
```
bartoszmajsak, line 48: nit: there's no need to remove "c" app from East cluster, you don't deploy it there, do you?

bartoszmajsak (commented):
I left few suggestions for you to consider.

cmd/federation-controller/main.go
bartoszmajsak: Fair enough

cmd/federation-controller/main.go
bartoszmajsak: I know, it's me. There's PR for that #103 so I won't mess it up again :)

internal/pkg/istio/config_factory_test.go
bartoszmajsak: I was actually thinking about bringing some assertion library and use `jq` for more readable tests. But I don't have a good example yet to show value in improved readability so I will ponder more about it.

cmd/federation-controller/main.go
bartoszmajsak: I think it's some leftover from me porting multi-peer WIP stuff, I also didn't like it when I noticed it now. So we are on the same page here.

bartoszmajsak: I think what you described is called "cuddling" :)

internal/pkg/istio/config_factory.go
jewertow: Thanks, I fixed this comment.

internal/pkg/istio/config_factory_test.go
jewertow: Yes, this was just a typo. Fixed

internal/pkg/istio/config_factory_test.go
jewertow: Copy-paste... I changed this name, I hope it will be clear now.

cmd/federation-controller/main.go
jewertow: Good idea, but I splitted functions for port, service name and service fqdn. Please look if that is ok?

internal/pkg/networking/resolve.go
jewertow: I did it unintentionally. I changed `fmt.Printf` to `log.Errorf` for now, but we will need better solution anyway. I will make another PR with more optimal solution to handling DNS changes.

test/e2e/remote_dns_name/main_test.go
jewertow: The godoc of `testing.M` says: "M is a type passed to a TestMain function to run the actual tests." Isn't it clear that `main_test.go` == `TestMain`?

test/e2e/remote_dns_name/main_test.go
jewertow: This function creates Deployments only in the west cluster, but Services are created in all clusters.
```
Setup(common.DeployApps(&common.WestApps, common.WestClusterName, namespace.Future(&common.AppNs), false, "b", "c")).
```
Istio test framework was designed to support pure multi-cluster scenarios, so I guess this is not a bug, but a feature.

test/e2e/remote_dns_name/main_test.go
bartoszmajsak: Coming from the Ginkgo world I missed it. Not a huge fan of this convention, but it is what it is. I rest my case.

test/e2e/remote_dns_name/main_test.go
bartoszmajsak: I need to dig deeper to better understand how it works under the hood. For the novice like me it reads a bit counter intuitive.

internal/pkg/config/federation_config.go
```diff
@@ -45,6 +47,18 @@ type Remote struct {
 	Network     string      `json:"network"`
 }
 
+func (r *Remote) ServiceName() string {
+	return fmt.Sprintf("federation-discovery-service-%s", r.Name)
+}
+
+func (r *Remote) ServiceFQDN() string {
+	return fmt.Sprintf("federation-discovery-service-%s.istio-system.svc.cluster.local", r.Name)
```
bartoszmajsak, line 55: Why not:

```suggestion
	return fmt.Sprintf("%s.istio-system.svc.cluster.local", r.ServiceName())
```

bartoszmajsak (changes requested):

internal/pkg/config/federation_config.go
bartoszmajsak: I also wonder if the namespace shouldn't be resolved dynamically here.

jewertow: > Why not:

Done

> I also wonder if the namespace shouldn't be resolved dynamically here.

We don't know the remote namespace for now, but we will be able to change it when we will have `remote.identity` implemented.

bartoszmajsak (approved):

test/e2e/common/setup.go
```diff
@@ -228,18 +233,22 @@ func InstallOrUpgradeFederationControllers(spireEnabled bool) resource.SetupFn {
 			if err != nil {
 				return err
 			}
+			remoteAddr := gatewayIP
+			if opts.SetRemoteDNSName {
+				remoteAddr = fmt.Sprintf("ingress.%s", remoteClusterName)
+			}
```
eoinfennessy, line 239: Minor readability suggestion avoiding variable reassignment:
```suggestion
			var remoteAddr string
			if opts.SetRemoteDNSName {
				remoteAddr = fmt.Sprintf("ingress.%s", remoteClusterName)
			} else {
			    remoteAddr = gatewayIP
			}
```

eoinfennessy (approved):
Looks good. I just added one nit.

test/e2e/common/setup.go
jewertow: Done :)

PR #94, opened 2024-11-29
chore(kind): simplifies cluster setup

bartoszmajsak:
This change expands the cluster provisioning logic to support more than two clusters.

Functions for creating clusters and MetalLB are simplified, allowing parameterized invocation from the main script.

Pod and service subnet parameters are removed as they are no longer required. See [this comment](https://github.com/openshift-service-mesh/federation/pull/92#discussion_r1862738653) for details.

Functions are updated to be POSIX-compliant, improving portability across different shells.

Additionally, the `upload_image` function is moved to `cluster.sh` where it belongs.

### DIFF
--- a/test/scripts/cluster.sh
+++ b/test/scripts/cluster.sh
@@ -2,60 +2,29 @@
 
 ROOT=$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)
 
-function create_kind_cluster {
-  cluster=$1
-  podSubnet=$2
-  svcSubnet=$3
-
-  echo "Creating cluster '$cluster' in podSubnet='$podSubnet' and svcSubnet='$svcSubnet'"
+provision_kind_cluster() {
+  local cluster=$1
 
   kind create cluster --name "$cluster" --config=<<EOF
 apiVersion: kind.x-k8s.io/v1alpha4
 kind: Cluster
-networking:
-  podSubnet: "$podSubnet"
-  serviceSubnet: "$svcSubnet"
 EOF
+
+  kind get kubeconfig --name "${cluster}" > $ROOT/${cluster}.kubeconfig
+
+  retry install_metallb $cluster
 }
 
-function provision_kind_clusters {
-  echo "Creating KinD clusters"
-  kind_pids=()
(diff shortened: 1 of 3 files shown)


test/scripts/cluster.sh
```diff
@@ -2,60 +2,29 @@
...
+provision_kind_cluster() {
+  local cluster=$1
 
   kind create cluster --name "$cluster" --config=<<EOF
 apiVersion: kind.x-k8s.io/v1alpha4
 kind: Cluster
-networking:
-  podSubnet: "$podSubnet"
-  serviceSubnet: "$svcSubnet"
 EOF
```
jewertow, line 11: Why not just this:
```suggestion
  kind create cluster --name "$cluster"
```

jewertow (approved):
Has `x()` any advantage over `function x` in bash?

jewertow:
> Has x() any advantage over function x in bash?

Nevermind... I didn't notice "Functions are updated to be POSIX-compliant, improving portability across different shells."

test/scripts/kind_provisioner.sh
```diff
@@ -9,22 +9,33 @@ set -x
...
+
+  pids=()
+  for cluster in "${clusters[@]}"; do
+    provision_kind_cluster "${cluster}" &
+    pids+=($!)
+  done
+
+  for pid in "${pids[@]}"; do
+    wait "$pid" || echo "Process $pid failed"
+  done
```
eoinfennessy, line 30: Should this exit with a non-zero status if the process fails?

eoinfennessy (approved):

test/scripts/cluster.sh
bartoszmajsak: True, we don't need this at all. Thanks for spotting it. Fixed in 92db29b

test/scripts/kind_provisioner.sh
bartoszmajsak: Yeah, we could. If any of the cluster provisioning fails, we need to recreate the whole setup regardless, based on the current logic that this refactoring preserved.

Fixed in a7594b4

PR #95, opened 2024-12-02
fix(make): ensures binary is built against up-to-date code

bartoszmajsak:
In order to ensure that up-to-date codebase is used to build the binary, we should automate  Golang stubs generation based on protobuf API changes.

This step was missing in the build process, requiring developer to call it explicitly.

With this fix, `proto` stubs are generated automatically. Additionally license headers are added, otherwise generation makes them disappear. As a bonus we ensure all relevant source files have now license headers added for each invocation of `build` target.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@ help:
 
 EXTRA_BUILD_ARGS?=
 .PHONY: build
-build: ## Builds the project
+build: proto add-license ## Builds the project
 	go get $(PROJECT_DIR)/...
 	go build -C $(PROJECT_DIR)/cmd/federation-controller -o $(PROJECT_DIR)/$(OUT_DIR)/federation-controller $(EXTRA_BUILD_ARGS)
 


Makefile
```diff
@@ -14,7 +14,7 @@ help:
 
 EXTRA_BUILD_ARGS?=
 .PHONY: build
-build: ## Builds the project
+build: proto add-license ## Builds the project
```
eoinfennessy, line 17: Instead of having `add-license` as a dependency here we could instead append `$(MAKE) add-license` to the commands in the `proto` target. This would ensure that the license headers are not removed when running `make proto` on its own.

eoinfennessy (approved):
Looks good. I added one suggestion.

jewertow (approved):

Makefile
bartoszmajsak: That's a good point. After thinking about it a bit more I would rather stick to the current approach.

I am looking at those two targets as separate, smaller units of work which you could "pipe".

Unless there's a way to supply a header (which I haven't looked at yet) similarly to what `controller-gen` exposes through `object:headerFile` I would prefer to stick to this composition. 

Especially if we add some more dependent targets to the `build` which might need adding license as well.

PR #99, opened 2024-12-04
feat(script):  subnet calucations supports multiple clusters 

bartoszmajsak:
Previous implementation was tailored to support only two subnets for `east` and `west` clusters.

This change simplifies sub-net calculation and makes it capable of calculating sub-nets for more than 2 clusters.


test/scripts/lib.sh
```diff
@@ -1,17 +1,18 @@
 #!/bin/bash
 
 retry() {
+  local command=$@
   local n=1
   local max=5
-  local delay=5
+  local delay=10
```
bartoszmajsak, line 7: @jewertow Not sure that is significant change. I noticed that more often than not I have two failed attempts for installing MetalLB, so I bumped retry to 10s based on scientific measures. However, new evidence suggests `8s` would be more accurate to avoid quantum entanglement.

test/scripts/find_subnets.py
```diff
@@ -0,0 +1,47 @@
+import ipaddress
+import argparse
+import math
+import sys
+
+def find_subnets(network, required_subnets):
+    try:
+        # Create additional subnets to avoid overlapping
+        required_subnets += 2
```
jewertow, line 9: We never needed to leave additional pool of addresses to avoid overlapping, and we have this weird logic only because I copied and reworked scripts from Istio, because I wanted to achieve similar multi-network setup. The original script was taking a subnet and allocated 100 (?) last addresses "to avoid overlapping". Integrating Istio test framework was complicated enough to forget about investigating why the original author claim that addresses may overlap. Now, we can remove it, because the default KinD installation does not create any LB service, and even if the cluster was provisioned with a LB service, it wouldn't have assigned any IP until MetalLB is installed, so there is no risk of overlapping IP addresses.
```suggestion
        # Create additional subnets to avoid overlapping
        required_subnets += 2
```

jewertow (changes requested):

test/scripts/find_subnets.py
bartoszmajsak: > Now, we can remove it

I am a bit puzzled by your suggestion then :)

bartoszmajsak: Anyway, thanks for the background story. Fixed in 89affcd

jewertow (approved):

PR #100, opened 2024-12-04
feat(fds): groups federated services by remote peers

### LINKED ISSUE #62: Identify source of discovery response

bartoszmajsak:
> [!IMPORTANT]
> This PR doesn't include the code required to handle multiple peers yet.
> This is a refactoring of existing data structures first and preserves the current logic.

This PR refactors existing code to facilitate more than one peer by keeping
each federated service (imported from remote) grouped by remote's name.

Initial assumption to rely on `XFCC` header was unsuccessful, as received gRPC 
stream did not have expected headers (or I couldn't get them by using gRPC API
for fetching Metadata in case of Aggregated Discover Service).

Instead of relying on the peer identity passed as part of discovery
response we can group federated services by remote names (aliases)
defined as initial config. 

With this approach we can ensure uniqueness of peer names locally
(which will be solved when they become Custom Resources anyway)
and avoid naming collisions which can lead to wrongly assigned 
federated services.

Fixes #62
Closes #101

### DIFF
--- a/internal/pkg/fds/imported_service_store.go
+++ b/internal/pkg/fds/imported_service_store.go
@@ -17,36 +17,44 @@ package fds
 import (
 	"sync"
 
+	"github.com/openshift-service-mesh/federation/internal/pkg/config"
+
 	"github.com/openshift-service-mesh/federation/internal/api/federation/v1alpha1"
 )
 
 // ImportedServiceStore is a thread-safe wrapper for current state of imported services
 type ImportedServiceStore struct {
 	mu               sync.RWMutex
-	importedServices []*v1alpha1.ExportedService
+	importedServices map[string][]*v1alpha1.ExportedService
 }
 
 func NewImportedServiceStore() *ImportedServiceStore {
 	return &ImportedServiceStore{}
 }
 
-func (s *ImportedServiceStore) Update(importedServices []*v1alpha1.ExportedService) {
+func (s *ImportedServiceStore) Update(importedServices map[string][]*v1alpha1.ExportedService) {
 	s.mu.Lock()
 	defer s.mu.Unlock()
-	newImportedServices := make([]*v1alpha1.ExportedService, 0, len(importedServices))
-	for _, svc := range importedServices {
-		newImportedServices = append(newImportedServices, svc.DeepCopy())
+
+	newImportedServices := make(map[string][]*v1alpha1.ExportedService)
+	for source, services := range importedServices {
+		for _, svc := range services {
+			newImportedServices[source] = append(newImportedServices[source], svc.DeepCopy())
+		}
 	}
+
 	s.importedServices = newImportedServices
 }
 
-// GetAll returns copy of the current state
-func (s *ImportedServiceStore) GetAll() []*v1alpha1.ExportedService {
+// From returns copy of all services exported from given remote peer.
+func (s *ImportedServiceStore) From(remote config.Remote) []*v1alpha1.ExportedService {
 	s.mu.RLock()
 	defer s.mu.RUnlock()
+
 	out := make([]*v1alpha1.ExportedService, 0, len(s.importedServices))
-	for _, svc := range s.importedServices {
+	for _, svc := range s.importedServices[remote.Name] {
 		out = append(out, svc.DeepCopy())
 	}
+
 	return out
 }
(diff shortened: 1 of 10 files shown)


internal/pkg/fds/imported_service_store.go
```diff
@@ -15,38 +15,47 @@
 package fds
 
 import (
+	"github.com/openshift-service-mesh/federation/internal/pkg/config"
 	"sync"
 
 	"github.com/openshift-service-mesh/federation/internal/api/federation/v1alpha1"
 )
 
+type FederatedSvcByPeer map[string][]*v1alpha1.ExportedService
```
bartoszmajsak, line 24: I introduced this alias type for convenience.

jewertow (changes requested):
We know the source, before we send a discovery request, so sending source over the network is redundant. I expected something like this: https://github.com/openshift-service-mesh/federation/pull/101.

bartoszmajsak:
> We know the source, before we send a discovery request, so sending source over the network is redundant. I expected something like this: #101.

In fact when I was reviewing this PR before opening it, I thought about this approach as well.

I concluded it wouldn't really be identifying the source of the `ExportedService` (like we would have if `spiffeid` was available) but more grouping responses by "client-local" alias of the other peer. Having it as part of the response might also eliminate a need of having `remote.name` at all as we could dynamically create needed resource when getting first response from the peer. If increasing the size of the payload is of a concern we could stick to client-local approach as you suggested.

eoinfennessy (commented):
What happens if two or more remote peers are given the same `cfg.MeshPeers.Local.Name`? In this case wouldn't the services exported by each be merged into the same value in `FederatedSvcByPeer`?

For this reason I also prefer the approach in #101 because we can ensure uniqueness of remote peer names using cross-field validation.

bartoszmajsak:
> What happens if two or more remote peers are given the same cfg.MeshPeers.Local.Name? In this case wouldn't the services exported by each be merged into the same value in FederatedSvcByPeer?

I would say that is naming collision and wrong configuration (and would also be present if we would go spiffeid). It's not straightforward to validate in contrast to checking if remote names/aliases are the unique per controller instance.  We would need to have some centralized registry for that and that seems to be a bit too much for what we need. 

I will rework this to use client aliases for remote similarly to #101.

bartoszmajsak:
> No, remote.name must be configured locally, because it's used to create unique ServiceEntries for remote FDS servers

I see, I missed the fact that we do not always rely on `remote.Addresses[0]` anymore and I was thinking name/detailed identity could come as a response from remote FDS when establishing a connection and passing initial discovery requests. And only then we could set up required resources based on the information we received. 

Without it defined as initial config we would end up with chicken-egg problem not knowing the name which is used for creating the service and thus calling it from the FDS client :)

internal/pkg/fds/imported_service_store.go
```diff
@@ -15,38 +15,47 @@
 package fds
 
 import (
+	"github.com/openshift-service-mesh/federation/internal/pkg/config"
```
jewertow, line 18: `make fix-imports`

internal/pkg/fds/imported_service_store.go
jewertow: TBH I would prefer map, but does not really matter.

jewertow (approved):

internal/pkg/fds/imported_service_store.go
bartoszmajsak: Ok, I will automate it sooner :) We shouldn't spend time on such obvious things :)

internal/pkg/fds/imported_service_store.go
bartoszmajsak: I just noticed it's all over the place, so I thought it can be handy. But to be honest it might be more noise than help as it pretends to be more than just a map and does not encapsulate any extra behaviour. I will revert it before merging

PR #102, opened 2024-12-05
feat(config): controller's namespace autodiscovery

bartoszmajsak:
We do not need special environment variable to determine where our controller runs. To achieve this we could either use Downward API environment variable called `POD_NAMESPACE` or rely on the fact that we are using Service Account which automatically propagates namespace as mounted secret.

This PR introduces `config.Namespace` function which resolves namespace in the exact order mentioned above or falls back to `istio-system`.

Bonus: it also re-uses Istio's underyling k8s client for the informers instead of creating new instance.

### DIFF
--- a/cmd/federation-controller/main.go
+++ b/cmd/federation-controller/main.go
@@ -25,16 +25,6 @@ import (
 	"time"
 
 	discovery "github.com/envoyproxy/go-control-plane/envoy/service/discovery/v3"
-	routev1client "github.com/openshift/client-go/route/clientset/versioned"
-	istiokube "istio.io/istio/pkg/kube"
-	istiolog "istio.io/istio/pkg/log"
-	"istio.io/istio/pkg/slices"
-	corev1 "k8s.io/api/core/v1"
-	"k8s.io/client-go/informers"
-	"k8s.io/client-go/kubernetes"
-	"k8s.io/client-go/rest"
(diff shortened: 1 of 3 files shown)


cmd/federation-controller/main.go
```diff
@@ -45,6 +35,13 @@ import (
 	"github.com/openshift-service-mesh/federation/internal/pkg/xds"
 	"github.com/openshift-service-mesh/federation/internal/pkg/xds/adsc"
 	"github.com/openshift-service-mesh/federation/internal/pkg/xds/adss"
+	routev1client "github.com/openshift/client-go/route/clientset/versioned"
+	istiokube "istio.io/istio/pkg/kube"
+	istiolog "istio.io/istio/pkg/log"
+	"istio.io/istio/pkg/slices"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/rest"
```
jewertow, line 44: It seems that goimports does not work as expected...

jewertow (approved):

cmd/federation-controller/main.go
bartoszmajsak: That's Goland sorting it alphabetically. I think that's better this way. I will eventually automate it, but no time now :)

PR #110, opened 2024-12-17
feat: introduces support for multiple peers

### LINKED ISSUE #51: Connect more than 2 peers
Our API allows to configure only 1 remote peer, but we want to connect more meshes. However, we must know the identity of the configured peers to determine where an instance of an exported service come from, so we must also extend the `Remote` struct with the identity that will include trust domain, namespace, and service account. These 3 information are needed to verify SPIFFE ID.

The API should look as follows:
```
federation:
  meshPeers:
    local:
      ...
    remote:
    - identity:
        trustDomain: cluster.local
        namespace: istio-system
        serviceAccount: federation-controller-west
      addresses:
      - 192.168.0.1
    - identity:
        trustDomain: cluster.local
        namespace: istio-system
        serviceAccount: federation-controller-east
    - addresses:
      - 192.168.1.1
```
Alternatively, the identity could be a single field containing the SPIFFE ID.
```
federation:
  meshPeers:
    local:
      ...
    remote:
    - identity: cluster.local/ns/istio-system/sa/federation-controller-east
      addresses:
      - 192.168.0.1
    - identity: cluster.local/ns/istio-system/sa/federation-controller-west
      addresses:
      - 192.168.1.1
```

The second option is more compact, but an average Istio user does not have to know the SPIFFE ID format, and additionally, the first option allows us to assume default values, e.g. the configuration below could result in the following SPIFFE ID: `cluster.local/ns/istio-system/sa/federation-controller-east`.
```
federation:
  meshPeers:
    remote:
    - identity:
        serviceAccount: federation-controller-east
```

Note that the identities must be unique, so when we connect meshes using the same trust domain, then federation controllers must have unique service account or namespace.

Server can determine the identity from the XFCC header set by the ingress gateway.

**Related issues:**

- Identify source peer (see https://github.com/openshift-service-mesh/federation/issues/62)

bartoszmajsak:
This PR adds support for multiple remote peers.  

### Core changes  

- Extends Federation structs and relevant CR creation to handle multiple remotes per controller.  
- Simplifies FDS server and client code.  
- Cleans up bootstrapping logic to support multiple remotes.  
- Adjusts service import logic.  

### Testing enhancements  

Most code changes in this PR are to enable multi-peer scenarios in our test setup:  

- Reworks test fixture definitions and to simplify multi-peer scenarios.  
- Groups test scenarios into a dedicated folder for better separation and easier navigation.  
- Introduces `East`, `Central`, and `West` clusters instead of name and index lookups, providing a wrapper around the Istio Test Framework `cluster.Cluster` struct.  
- Defines the `DeployOption` interface to decouple configuration changes from the main deploy logic, allowing each test scenario granular control over application setup.  

Fixes #51


internal/pkg/istio/config_factory.go
```diff
@@ -63,54 +63,60 @@ func NewConfigFactory(
...
+				Name:      fmt.Sprintf("mtls-sni-%s-%s", svcName, svcNs),
+				Namespace: cf.cfg.MeshPeers.Local.ControlPlane.Namespace,
+				Labels:    map[string]string{"federation.istio-ecosystem.io/peer": "todo"},
+			}
+		}
 
-	for _, svc := range cf.importedServiceStore.From(remote) {
-		dr := &v1alpha3.DestinationRule{
-			ObjectMeta: createObjectMeta(svc.Name, svc.Namespace),
+		drPerRemote := []*v1alpha3.DestinationRule{{
```
bartoszmajsak, line 82: Need to review this logic, does not sound right looking at right now

test/e2e/setup/federation.go
```diff
@@ -0,0 +1,48 @@
...
+)
+
+func InstallOrUpgradeFederationControllers(options ...CtrlOption) resource.SetupFn {
+	return func(ctx resource.Context) error {
+		for _, c := range ctx.Clusters() {
+			localCluster := Resolve(c)
+			remoteClusters := ctx.Clusters().Exclude(localCluster)
+			if out, err := localCluster.ConfigureFederationCtrl(remoteClusters, options...); err != nil {
+				return fmt.Errorf("failed to upgrade federation controller (cluster=%s): %v, %w", localCluster.ContextName, string(out), err)
+			}
```
bartoszmajsak, line 34: Can run it in go routines.

internal/pkg/istio/config_factory.go
```diff
@@ -63,54 +63,60 @@ func NewConfigFactory(
...
-				Tls: &istionetv1alpha3.ClientTLSSettings{
-					Mode: istionetv1alpha3.ClientTLSSettings_ISTIO_MUTUAL,
-					Sni:  routerCompatibleSNI(remote.ServiceName(), "istio-system", remote.ServicePort()),
-				},
-			},
-		},
-	}}
+		createObjectMeta := func(svcName, svcNs string) metav1.ObjectMeta {
+			return metav1.ObjectMeta{
+				Name:      fmt.Sprintf("mtls-sni-%s-%s", svcName, svcNs),
```
jewertow, line 76: We must append peer name to avoid overwriting resources when 3+ meshes are connected and 2 of them export the same service (for HA) and default cluster domain is used.

internal/pkg/istio/config_factory.go
```diff
@@ -222,97 +228,120 @@ func (cf *ConfigFactory) EnvoyFilters() []*v1alpha3.EnvoyFilter {
...
+		}
+
+		for _, importedSvc := range cf.importedServiceStore.From(remote) {
+			_, err := cf.serviceLister.Services(importedSvc.Namespace).Get(importedSvc.Name)
+			if err != nil {
+				if !errors.IsNotFound(err) {
+					return nil, fmt.Errorf("failed to get Service %s/%s: %w", importedSvc.Name, importedSvc.Namespace, err)
+				}
+				// Service doesn't exist - create ServiceEntry.
+				// TODO(multi-peer) handle naming clash & different resolution strategy (known limitation?)
```
jewertow, line 256: We are not able to properly import a service from multiple meshes when different resolution strategies are required and the default cluster domain is used. I guess that the majority of users will prefer custom domains for imported services, and then this problem will not occur. We will document this as a known issue.

internal/pkg/istio/config_factory.go
```diff
@@ -222,97 +228,120 @@ func (cf *ConfigFactory) EnvoyFilters() []*v1alpha3.EnvoyFilter {
...
+				// What if we happen to have multiple remotes with exported serviceName/serviceNamespace but:
+				// - each remote has different address defined (leading to static or DNS resolution)
+				// - services per remote peer differ on service ports
```
jewertow, line 259: This is very interesting case, and we can't handle it. It shows how important it is to support custom domains ASAP.

internal/pkg/istio/config_factory.go
```diff
@@ -222,97 +228,120 @@ func (cf *ConfigFactory) EnvoyFilters() []*v1alpha3.EnvoyFilter {
 }
 
 func (cf *ConfigFactory) ServiceEntries() ([]*v1alpha3.ServiceEntry, error) {
-	remote := cf.cfg.MeshPeers.Remote
 
-	if len(remote.Addresses) == 0 {
-		return nil, nil
-	}
+	var serviceEntries []*v1alpha3.ServiceEntry
+	serviceEntriesByHost := make(map[string]*v1alpha3.ServiceEntry, len(cf.cfg.MeshPeers.Remotes))
```
jewertow, line 233: I don't understand the purpose of this map. You never add entries to this map.

test/e2e/scenarios/remote_dns_name/main_test.go
```diff
@@ -0,0 +1,64 @@
...
+		Setup(setup.RecreateControlPlaneNamespace).
+		Setup(setup.CreateCACertsSecret).
+		Setup(setup.DeployControlPlanes("k8s")).
+		Setup(coredns.PatchHosts).
+		Setup(setup.InstallOrUpgradeFederationControllers(setup.RemoteAddressDNSName{})).
+		Setup(namespace.Setup(&setup.Namespace, namespace.Config{Prefix: "app", Inject: true})).
+		// a - client
+		// b - service available in east and west clusters - covers importing with WorkloadEntry
+		// c - service available only in west cluster - covers importing with ServiceEntry
+		Setup(setup.Clusters.East.DeployEcho(namespace.Future(&setup.Namespace), "a", setup.WithAllPorts{})).
```
jewertow, line 46: If the last argument is a var args, couldn't we skip `setup.WithAllPorts{}`?

test/test_utils.go
jewertow, line 1: Is this really better than this?
```
_, file, _, _ = runtime.Caller(0)
RootDir       = filepath.Join(filepath.Dir(file), "../../..")
```
You only needed to append additional "/.."

test/e2e/scenarios/remote_dns_name/main_test.go
bartoszmajsak: We could add it as default if empty

test/test_utils.go
bartoszmajsak: This way it's refactoring friendly if you move this file around. Reused my old solution

internal/pkg/istio/config_factory.go
bartoszmajsak: Yeah, noticed it this morning 
... Silly partial git add or some other brain fart... Must have been tired or stupid. Most likely both xD

test/e2e/scenarios/remote_dns_name/main_test.go
bartoszmajsak: Or even better - always add it. It replaces part of struct anyway

internal/pkg/istio/config_factory.go
jewertow: I wasn't right. We can solve it. If the same service is imported from clusters having different resolution strategies, we just need to set STATIC resolution and resolve addresses being DNS names. But I think you don't have to solve this issue in this PR. You can leave this TODO, and we can solve it later in a follow-up PR - that will be even easier to review.

internal/pkg/istio/config_factory.go
bartoszmajsak: This has been fixed with the latest changes pushed just now. PTAL at [this comment](https://github.com/openshift-service-mesh/federation/pull/110/files#diff-c1818d5fcdc653f21b85a94acbfa2fdafb66928e2610bcfb8d9cc32001572caeR287-R288) where I explained my line of thinking

test/e2e/scenarios/remote_dns_name/main_test.go
bartoszmajsak: That's reworked now, thanks for the idea!

bartoszmajsak:
Spire test suite is failing at the moment - I will investigate further.

cmd/federation-controller/main.go
```diff
@@ -89,6 +89,11 @@ func main() {
 		log.Fatalf("failed to parse configuration passed to the program arguments: %v", err)
 	}
 
+	// TODO(multi-peer): if 0 addresses given peer is not interested in importing anything - rework/rethink validation
```
bartoszmajsak, line 92: - [ ] Need to rework this

internal/pkg/istio/config_factory.go
```diff
@@ -63,54 +63,63 @@ func NewConfigFactory(
...
-	}
-
-	createObjectMeta := func(svcName, svcNs string) metav1.ObjectMeta {
-		return metav1.ObjectMeta{
-			Name:      fmt.Sprintf("mtls-sni-%s-%s", svcName, svcNs),
-			Namespace: cf.cfg.MeshPeers.Local.ControlPlane.Namespace,
-			Labels:    map[string]string{"federation.istio-ecosystem.io/peer": "todo"},
+	for _, remote := range cf.cfg.MeshPeers.Remotes {
+		if remote.IngressType != config.OpenShiftRouter {
+			return nil
```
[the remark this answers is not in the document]
bartoszmajsak: I think I overlooked it.. It's a condition for a single remote as it is now on the main branch. I think originally I wanted to skip it with log (`continue`) but you're right, we should document it as a limitation for now.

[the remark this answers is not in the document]
bartoszmajsak: Changed in 5cf8282

internal/pkg/istio/config_factory.go
bartoszmajsak: Done in 9dc6296

test/e2e/setup/federation.go
bartoszmajsak: 3a07c99

jewertow (changes requested):

cmd/federation-controller/main.go
bartoszmajsak: Correct, I have it removed already but it's not pushed yet.

chart/templates/clusterrole.yaml
```diff
@@ -12,7 +12,7 @@ rules:
 - apiGroups: ["security.istio.io"]
   resources: ["peerauthentications"]
   verbs: ["get", "list", "create", "update", "patch", "delete"]
-{{- if eq .Values.federation.meshPeers.remote.ingressType "openshift-router" }}
+{{- if (include "remotes.hasOpenshiftRouterPeer" .) }}
```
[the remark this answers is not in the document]
bartoszmajsak: Yeah, we are going make it dynamic now. This part though is rather easy - we can just grant those roles for `destinationrules` regardless.

For the remaining part - specifically adding role for the `routes/custom-host` we could check if we are running ctrl in openshift when reconciling and create the necessary role then.

Either way - we should capture it as sub-task for CRD story.

test/e2e/scenarios/remote_dns_name/main_test.go
```diff
@@ -0,0 +1,64 @@
...
+		Setup(setup.Clusters.East.DeployEcho(namespace.Future(&setup.Namespace), "a")).
+		Setup(setup.Clusters.East.DeployEcho(namespace.Future(&setup.Namespace), "b")).
+		Setup(setup.Clusters.West.DeployEcho(namespace.Future(&setup.Namespace), "b")).
+		Setup(setup.Clusters.West.DeployEcho(namespace.Future(&setup.Namespace), "c")).
+		Setup(setup.Clusters.Central.DeployEcho(namespace.Future(&setup.Namespace), "b")).
+		Setup(setup.Clusters.Central.DeployEcho(namespace.Future(&setup.Namespace), "d")).
+		// c and d must be removed from other clusters, because we want to test importing a service
+		// that exists only in the remote cluster.
+		Setup(setup.RemoveServiceFromClusters("c", namespace.Future(&setup.Namespace), &setup.Clusters.East, &setup.Clusters.Central)).
+		Setup(setup.RemoveServiceFromClusters("d", namespace.Future(&setup.Namespace), &setup.Clusters.East, &setup.Clusters.West)).
```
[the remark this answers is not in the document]
bartoszmajsak: The simpler, the better. I reworked it only to have 3 services as you suggested and moved the common setup to a single function so we can consistently re-use them in all scenarios.

However, we should be only removing "c" Service from East cluster with this setup.

Added in c069418

internal/pkg/istio/config_factory.go
bartoszmajsak: #123

@jewertow PTAL, and check if I captured what you have in mind accurately

jewertow (approved):

PR #111, opened 2024-12-18
fix(deps): pins protobuf versions

bartoszmajsak:
To ensure reproducible builds and avoid accidental changes in unrelated
PRs caused by differences in local binary versions (and so code generation), we should pin the versions of the tools we use.


bartoszmajsak:
Next step would be to add those deps as part of go.mod and configure dependabot to manage updates for us. WDYT @jewertow ?

jewertow:
> WDYT @jewertow ?

Sounds good.

jewertow (approved):

PR #116, opened 2024-12-19
feat(make): generates code only on API changes

bartoszmajsak:
Instead of relying on phony `proto` target we can leverage `make`s ability to trigger code generation only when source (definition of an API) has been modified.

This PR reworks proto generation target to only remake generated `*.go` files when source `*.proto` files are newer (based on mtime check).


Makefile
```diff
@@ -118,13 +123,9 @@ clean:
...
 
-PROTO_DIR=$(PROJECT_DIR)/api/proto/federation
-API_GEN_DIR=$(PROJECT_DIR)/internal/api
-
-.PHONY: proto
-proto: $(PROTOC) $(PROTOC_GEN_GO) $(PROTOC_GEN_GRPC) $(PROTOC_GEN_DEEPCOPY) ## Generates Go files from protobuf-based API files
-	@PATH=$(LOCALBIN):$$PATH $(PROTOC) --proto_path=$(PROTO_DIR) --go_out=$(API_GEN_DIR) --go-grpc_out=$(API_GEN_DIR) --golang-deepcopy_out=:$(API_GEN_DIR) $(PROTO_DIR)/**/*.proto
-
+$(PROTOBUF_GEN): $(PROTOBUF_API_SRC) $(PROTOC) $(PROTOC_GEN_GO) $(PROTOC_GEN_GRPC) $(PROTOC_GEN_DEEPCOPY) ## Generates Go files from protobuf-based API files
+	@echo $(PROTOBUF_API_DIR)
```
bartoszmajsak, line 127:
```suggestion
```

jewertow (approved):

PR #119, opened 2024-12-19
chore(docs): simplifies spire example

bartoszmajsak:
There are a few minor shortcomings how the current spire example is
structured.

Each time bundle propagation is executed it appends to the existing file
which is a template and under the source control. This PR reworks it to
apply changes in place.

In addition `trust-bundle-federation.yaml`s are places in a
counter-intuitive way, where the "east" one is in the "west" folder and
the other way around. If we add third cluster, we would end up with 6 of
those files, which in fact are templates. This PR changes it to have one
template with cluster name being replaced on the fly.

The <placeholders> have been renamed to ${VAR} so the files can be also
processed by tools like `envsubst` easily.

Helm installation of Spire has `--wait` flag now to prevent from
federating bundles too early which might lead to missing bundles in one
of the clusters.


examples/spire/README.md
```diff
@@ -86,12 +84,12 @@ WEST_GATEWAY_IP=$(kwest get svc federation-ingress-gateway -n istio-system -o js
 helm-east install east-mesh chart -n istio-system \
     --values examples/kind/east-federation-controller.yaml \
     --set "istio.spire.enabled=true" \
-    --set "federation.meshPeers.remote.addresses[0]=$WEST_GATEWAY_IP"
+    --set "federation.meshPeers.remote.addresses[0]=$WEST_GATEWAY_IP" \
```
jewertow, line 87: This backslash breaks this command.
```suggestion
    --set "federation.meshPeers.remote.addresses[0]=$WEST_GATEWAY_IP"
```

examples/spire/README.md
```diff
@@ -86,12 +84,12 @@ WEST_GATEWAY_IP=$(kwest get svc federation-ingress-gateway -n istio-system -o js
...
 EAST_GATEWAY_IP=$(keast get svc federation-ingress-gateway -n istio-system -o jsonpath='{.status.loadBalancer.ingress[0].ip}')
 helm-west install west-mesh chart -n istio-system \
     --values examples/kind/west-federation-controller.yaml \
     --set "istio.spire.enabled=true" \
-    --set "federation.meshPeers.remote.addresses[0]=$EAST_GATEWAY_IP"
+    --set "federation.meshPeers.remote.addresses[0]=$EAST_GATEWAY_IP" \
```
jewertow, line 92: Same here:
```suggestion
    --set "federation.meshPeers.remote.addresses[0]=$EAST_GATEWAY_IP"
```

jewertow (changes requested):

examples/spire/README.md
bartoszmajsak: Humm weird. I tested it yesterday and might have missed it.

jewertow:
> Helm installation of Spire has --wait flag

This makes it slower, because Spire installations are not processed in parallel, so that was intentional. But I always watch installation, so `--wait` makes sense when someone just goes through the steps and checks the final command.

bartoszmajsak:
> > Helm installation of Spire has --wait flag
> 
> This makes it slower, because Spire installations are not processed in parallel, so that was intentional. But I always watch installation, so `--wait` makes sense when someone just goes through the steps and checks the final command.

That was my motivation.

jewertow (approved):

PR #120, opened 2025-01-02
feat(design): CRDs and controllers

jewertow:
The main purpose of this design is to preserve the history and context of important decisions made for the project.

You can read a preview of this doc [here](https://github.com/jewertow/federation/blob/crds-and-kubebuilder-proposal/docs/enhancements/1-crds-and-controllers.md).


eoinfennessy (commented):
Looks like a solid idea. I like that it is essentially just two singleton resources that a user has to manage. I added a few thoughts regarding using `FederatedServicePolicy` for service import policies and specifying grouping policies by peer.

docs/enhancements/1-crds-and-kubebuilder.md
```diff
@@ -0,0 +1,290 @@
...
+  - meshID: west
+    status: Connected
+  - meshID: central
+    status: Disconnected
+    lastErrorMessage: "No route to host 192.168.2.1"
+```
+
+#### FederatedServicePolicy
+
+`FederatedServicePolicy` is a namespaced resource for specifying rules to export (and import - TBD) services.
```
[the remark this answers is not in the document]
eoinfennessy: > Why not 1 CM per each imported service?

This would work too. We'd just need a way (probably a label) to list all exported service CMs associated with a/all peer(s).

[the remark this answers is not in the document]
jewertow: 1 CM per 1 imported service will be better for delta FDS. The controller will be simpler, because we will reconcile only resources related to a single service.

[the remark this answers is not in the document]
jewertow: @eoinfennessy I've changed my mind about CMs. I would prefer ImportedService CRD, because:
1. We need a controller anyway - even if we used CMs, we would need a controller to reconcile user's changes.
2. Creating, potentially, dozens or hundreds of CMs in various namespaces does not sound like a good UX.
3. ServiceImport from MCS API is also invented to be used only by controllers, so maybe this approach is not bad.

docs/enhancements/1-crds-and-kubebuilder.md
```diff
@@ -0,0 +1,380 @@
...
+        ).
+        Watches(
+            &v1alpha1.MeshFederation{},
+            handler.EnqueueRequestsFromMapFunc(),
+        ).
+        Complete(r)
+}
+```
+
+`ImportedService` will be applied by FDS client like below:
```
jewertow, line 347: I think we could rely on `EventRecorder` to trigger reconciliation without the actual event in k8s, instead of applying objects. I will investigate this approach.

docs/enhancements/1-crds-and-kubebuilder.md
```diff
@@ -0,0 +1,380 @@
+| Status | Authors   | Created    | 
+|--------|-----------|------------|
+| Review | @jewertow | Jan 2 2025 |
+
+# CRDs and kubebuilder
+
+## Overview
+
+We need CRDs to properly manage lifecycle of the Istio resources created by our controller.
+Currently, resources are not reconciled on user changes and are not removed when the controller is uninstalled, and that results in mesh misconfiguration.
```
bartoszmajsak, line 10: Another benefit here is that we can easily federate more meshes by providing external configuration (through CR) that does not require ctrl pod to restart, does not need error-prone cli flags as the only source of config, we can attach status to it without a need to parse events or logs to name a few. This is quite a UX improvement in my eyes.

docs/enhancements/1-crds-and-kubebuilder.md
```diff
@@ -0,0 +1,380 @@
...
+      # If "openshift-router" is set the controller applies DestinationRules with SNI compatible with OpenShift Router.
+      # If "istio" is set client mTLS settings are not modified.
+      type: istio
+status:
+  conditions:
+  - meshID: west
+    status: Connected
+  - meshID: central
+    status: Disconnected
+    lastErrorMessage: "No route to host 192.168.2.1"
```
bartoszmajsak, line 97: I would stick to [typical condition properties](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties) which can be understood by external tooling.

docs/enhancements/1-crds-and-kubebuilder.md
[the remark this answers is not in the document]
eoinfennessy: The intent of my original comment was to suggest using one controller (`FederatedServicePolicy`) to manage both export and import configs. A ConfigMap is a poor choice for representing imported services for the reasons you have listed, but controller-less CRD (`ImportedService`) could instead be used, providing a better structure than a CM.

Why I'm suggesting this: An `ImportedService` controller will always depend on the config of a `FederatedServicePolicy` CR because that is where we are proposing the import rules will get specified. So why not have the `FederatedServicePolicy` controller also manage import configs that are associated with those rules instead of having the `ImportedService` controller depend on a `default` `FederatedServicePolicy` to determine how to reconcile each `ImportedService`?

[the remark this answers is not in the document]
jewertow: > An `ImportedService` controller will always depend on the config of a `FederatedServicePolicy` CR

No, `ImportedService` will be created for each imported service received by the FDS server. `FederatedServicePolicy` may limit importing services, but it does not specify what is actually imported. It will only specify the expected state, not the actual.

docs/enhancements/1-crds-and-kubebuilder.md
jewertow: Ok, I was looking for a workaround to trigger reconciliation without creating objects in k8s. We can try to trigger reconciliation using a channel.

docs/enhancements/1-crds-and-kubebuilder.md
```diff
@@ -0,0 +1,380 @@
...
+        - reviews
+    # Namespace selectors allows to export all services from namespaces selected by label 
+    namespaceSelectors:
+    - matchLabels:
+        istio-injection: enabled
+    # Service list allows to export particular services in particular namespaces.
+    serviceList:
+    - "ratings/ns-1"
+    - "reviews/ns-1"
+    - "*/ns-2"
```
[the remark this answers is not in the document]
jewertow: Yes, right. I fixed it, thanks!

docs/enhancements/1-crds-and-kubebuilder.md
```diff
@@ -0,0 +1,380 @@
...
+# TODO
+# status:
+```
+
+Rules based on label selectors are useful when an admin wants to give users the control on exporting Services,
+while `serviceList` gives the admin full control on exported services.
+
+We are able to provide the same functionality with only label selectors for services and namespaces.
+In this variant, `export` is a list of rule sets containing `serviceSelector` or `namespaceSelector`.
+Rules are OR-ed, but each rule set (an item in the `export` list) is AND-ed.
```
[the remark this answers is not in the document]
jewertow: Ok, thanks for sharing your opinion on that!

docs/enhancements/1-crds-and-kubebuilder.md
jewertow: I rephrased the overview and included these arguments, thanks.

docs/enhancements/1-crds-and-kubebuilder.md
jewertow: Thank you very much for this suggestion! I didn't know that convention, so I had no idea how to approach to this problem.

docs/enhancements/1-crds-and-kubebuilder.md
```diff
@@ -0,0 +1,290 @@
...
+  - meshID: central
+    status: Disconnected
+    lastErrorMessage: "No route to host 192.168.2.1"
+```
+
+#### FederatedServicePolicy
+
+`FederatedServicePolicy` is a namespaced resource for specifying rules to export (and import - TBD) services.
+This resource is expected to be created as a single instance for all exported (and imported - TBD) services.
+This is necessary, because all exported services will be associated with a single e/w gateway, so we can't map n to 1 resources.
```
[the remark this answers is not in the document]
jewertow: Please read the doc again, because I clarified export and import semantics.

docs/enhancements/1-crds-and-kubebuilder.md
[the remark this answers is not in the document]
jewertow: > Can we follow the template of [SEP](https://github.com/istio-ecosystem/sail-operator/blob/main/enhancements/SEP1-enhancement-process.md)? This will help to flesh it out even more.

I do not understand. If I followed that template, what would be fleshed out?

> clear explanation of why this is better than the current approach

I added more details to the overview, so it should be clear now.

docs/enhancements/1-crds-and-kubebuilder.md
```diff
@@ -0,0 +1,380 @@
...
+          name: tls-passthrough
+          # Optional.
+          # Port of the ingress gateway Service.
+          number: 15443
+    # Optional.
+    controlPlane:
+      # Optional.
+      # Control plane namespace used to create mesh-wide resources.
+      namespace: istio-system
+  remote:
```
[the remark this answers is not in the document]
jewertow: > That would for example make status reporting according to conventions

Agree, this would clarify the status. On the other hand, it complicates the installation. I don't know...

> but also any change to the remote config wouldn't require reconciling the whole thing, but just the relevant part

I think it does not matter, because reconciliation will result in updating shared configuration.

docs/enhancements/1-crds-and-kubebuilder.md
```diff
@@ -0,0 +1,380 @@
...
+
+`FederatedServicePolicy` is a namespaced resource for specifying rules to export (and import - TBD) services.
+This resource is expected to be created as a single instance for all exported (and imported - TBD) services.
+This is necessary, because all exported services will be associated with a single e/w gateway, so we can't map n to 1 resources.
+
+The key assumptions for export and import semantics in all variants of `FederationServicePolicy`:
+1. Export and import rules DO NOT enforce any authorization policy.
+2. Export and import rules ensure cross-cluster service discovery, and are not intended for enforcing security.
+3. Export rules are always federation-wide - we do not allow to export services to particular meshes in a federation.
+4. Import rules are defined per remote mesh to allow narrowing down service discovery data.
```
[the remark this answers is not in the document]
jewertow: > Policy can be conflated with authorization or any other concept which means something to be enforced

Right, thanks! I changed it to `FederationServiceRules`.

docs/enhancements/1-crds-and-kubebuilder.md
```diff
@@ -0,0 +1,380 @@
...
+    dnsSettings:
+      # Optional
+      # Search domain is a suffix for the service FQDN
+      searchDomain: mesh.global
+      # Optional
+      # If true, then a service will be exposed as <original-svc-name>.<original-svc-ns>.<search-domain>
+      includeNamespace: false
+      # Examples:
+      # 1. `searchDomain: svc.west.mesh` and `includeNamespace: true` will export ratings and reviews as ratings.bookinfo.svc.west.mesh and reviews.bookinfo.svc.west.mesh
+      # 2. `searchDomain: mesh.global` and `includeNamespace: false` will export ratings and reviews as ratings.mesh.global and reviews.mesh.global.
```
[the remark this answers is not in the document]
jewertow: In general, that makes sense to me and I would agree, but I think we will have to allow aliasing the service name. E.g. you have a service named `productpage-v1`, and you want to export it as `productpage.mesh.global`. This use case will require us to change `serviceList` from list of strings to list of objects like this:
```
serviceList:
- service: ns-1/reviews-v1
  alias: reviews
```
so then we will end up in having DNS settings in the `rules` section anyway. WDYT?

docs/enhancements/1-crds-and-controllers.md
```diff
@@ -0,0 +1,430 @@
...
+  - productpage.bookinfo.svc.mesh.global
+  ...
+```
+
+> [!IMPORTANT]
+> As the `ImportedService` is fully managed by the controller, the initial reconciliation must be triggered programmatically by the FDS client.
+> We should evaluate possible option and choose the best one during the implementation.
+> Example options:
+> * FDS client can call the controller function directly during processing FDS response.
+> * FDS client can trigger the reconciliation loop sending an imported service via a channel to the controller.
```
eoinfennessy, line 391: I don't understand this. Will the initial reconciliation not be triggered when the FDS client creates/applires the `ImportedService`?

docs/enhancements/1-crds-and-controllers.md
```diff
@@ -0,0 +1,430 @@
...
+
+#### Egress gateway controller
+
+Egress gateway can be used in routing traffic to imported services, but it cannot be owned by `ImportedService`,
+because all imported services will be handled by one `Gateway` and one `VirtualService`, so this is N:1 relation.
+Therefore, resources related to egress gateway will be handled by a dedicated controller, and their owner will be `FederationServiceRules`.
```
eoinfennessy, line 397: Why add an additional controller for egress gateways instead of making this an additional responsibility of the `FederationServiceRules` controller?

docs/enhancements/1-crds-and-controllers.md
```diff
@@ -0,0 +1,430 @@
+| Status | Authors   | Created    | 
+|--------|-----------|------------|
+| Review | @jewertow | Jan 2 2025 |
+
+# CRDs and controllers
+
+## Overview
+
+Currently, we use helm to manage the controller deployment and its configuration. The configuration is passed to the controller as CLI arguments.
+This approach requires restarting the controller on any configuration change and causes pod failures in case of a misconfiguration.
```
bartoszmajsak, line 10:
```suggestion
Currently, we use Helm to manage the controller deployment and its configuration, which is passed as CLI arguments. This approach offers limited flexibility, as any configuration change requires a controller restart and can lead to pod failures in case of misconfiguration.
```

docs/enhancements/1-crds-and-controllers.md
```diff
@@ -0,0 +1,430 @@
...
+
+Additionally, current approach to resource management is not reliable. We apply resources when it's necessary using the Istio client's `Apply` function,
+so we are not able to reconcile created objects on user's modifications or controller uninstallation.
```
bartoszmajsak, line 13:
```suggestion
We create required Istio resources only when necessary, driven exclusively by notifications from the FDS client. However, this approach has a drawback - it is not reliable and does not ensure the desired state of the cluster after resources are created. Since there is no ongoing reconciliation, any drift from the intended configuration, whether due to user modifications or controller uninstallation, remains uncorrected.
```

docs/enhancements/1-crds-and-controllers.md
```diff
@@ -0,0 +1,430 @@
...
+
+### Goals
+
+Design CRDs for better user experience and ensure proper resource lifecycle and garbage collection.
```
bartoszmajsak, line 17: I was thinking we can expand the goals based on what has been mentioned in the subsequent sections as well as overview, which is focused on drawbacks of existing approach.

```suggestion
We aim to improve the reliability and flexibility of Service Mesh Federation by:

- Introducing CRDs for a better user experience and declarative resource management.
- Ensuring proper resource lifecycle management.
- Preventing configuration drift by detecting and correcting unintended modifications.
```

docs/enhancements/1-crds-and-controllers.md
```diff
@@ -0,0 +1,430 @@
...
+
+### Non-goals
+
+Design CRD for installing federation controller. Federation controller will be still installed with Helm,
+and we are not planning any dedicated operator or integration with Sail Operator for now.
```
bartoszmajsak, line 22: IIUC publishing it as OLM bundle is not on the roadmap at all?

docs/enhancements/1-crds-and-kubebuilder.md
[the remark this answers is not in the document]
bartoszmajsak: I'm a bit torn on this.

A dedicated Custom Resource (such as `Peer` or something similar) offers the advantage of being self-contained, with its own status, while also allowing reconfiguration of only the relevant client.

On the other hand, managing everything under a single slice in `MeshFederation` simplifies user operations when reconfiguring multiple peers at once. However, this approach requires additional logic to determine whether a given client needs reconfiguration. We can blindly reconfigure everything though... not sure if that's a good approach to be honest.

docs/enhancements/1-crds-and-controllers.md
```diff
@@ -0,0 +1,430 @@
...
+```yaml
+apiGroup: federation.openshift-service-mesh.io/v1alpha1
+kind: ImportedService
+metadata:
+  # Name is a dash-separated FQDN of an imported service
+  name: productpage-bookinfo-svc-mesh-global
+  # Namespace depends on whether the service exists in the local cluster.
+  # If it does not exist locally, it will be root mesh namespace, otherwise it will be the original namespace. 
+  namespace: istio-system
+  # Having MeshFederation set as an owner ensures proper resource cleanup after deleting federation.
```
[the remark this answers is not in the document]
jewertow: > Where will the OwnerReference be set?

In ImportedService resources.

> When created by the FDS client? If so, the FDS client will need to get the MeshFederation resource to set the owner reference.

No, the FDS client will trigger reconciler without creating the object in k8s, e.g. send internal structure to the reconciler via a channel.

> I'd presume this would need to be done each time a new ImportedService is created?

Don't reconcilers use cached informers under the hood?

> An alternative to setting an owner could be to add a finalizer to the MeshFederation controller that cleans up all ImportedServices prior to the deletion of the default MeshFederation.

Thanks for this suggestions. I will think it.

docs/enhancements/1-crds-and-controllers.md
jewertow: Egress gateway and its virtual service are tightly coupled with imported services. `FederationServiceRules` includes selectors for imported services, but it won't be triggered on every service import.

docs/enhancements/1-crds-and-controllers.md
```diff
@@ -0,0 +1,430 @@
...
+      type: istio
+```
+
+The controller for this resource will be responsible only for updating the mesh configuration in other components,
+like Istio resource factory, FDS clients manager, etc. It will not manage any child resources.
+
+#### FederationServiceRules
+
+`FederationServiceRules` is a namespaced resource for specifying export/import rules.
+This resource is expected to be created as a single instance for all exported services.
```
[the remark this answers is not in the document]
jewertow: 1. We will start with an admission webhook to ensure that we have only one instance.
2. We may consider using labels to associate MeshFederation and FederationServiceRules in the future, e.g. if we will want to support multi-tenant deployments.

docs/enhancements/1-crds-and-controllers.md
```diff
@@ -0,0 +1,430 @@
...
+  - central
+```
+
+Responsibility of the controller for this resource:
+1. Owning import-related resources: `ServiceEntry` or `WorkloadEntry`, and `DestinationRule`.
+2. Restoring the resource state in case of user's modification.
+3. Watching services and recreating `ImportedServices` when it's necessary, example:
+    * t0: `ImportedService` for `productpage` service was created, and that service does not exist in the local cluster, so `ServiceEntry` in `istio-system` namespace was created.
+    * t1: User creates `productpage` service in namespace `bookinfo`.
+    * t2: Controller removes `ImportedService` from `istio-system` namespace with its child resources and creates it again in `bookinfo` namespace with proper child resources.
```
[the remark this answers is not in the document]
bartoszmajsak: If we decide not to use `ImportedService` as `ownerRef` of the child resources, we can keep it in the root ns (`istio-system`) and move only dependand resources (i.e. `WorkloadEntry`) to cover this case.

Having all internal resources such as `ImportedService` in a single ns can simplify managing them.

docs/enhancements/1-crds-and-controllers.md
bartoszmajsak: @jewertow It would be good to define a use-case when we need `egress` gateway and what determines its configuration. Having this will help to decide which controller should be responsible for configuring it.

docs/enhancements/1-crds-and-controllers.md
```diff
@@ -0,0 +1,430 @@
...
+      # Search domain is a suffix for the service FQDN
+      searchDomain: mesh.global
+      # Optional
+      # If true, then a service will be exposed as <original-svc-name>.<original-svc-ns>.<search-domain>
+      includeNamespace: false
+      # Examples:
+      # 1. `searchDomain: svc.west.mesh` and `includeNamespace: true` will export ratings and reviews as ratings.bookinfo.svc.west.mesh and reviews.bookinfo.svc.west.mesh
+      # 2. `searchDomain: mesh.global` and `includeNamespace: false` will export ratings and reviews as ratings.mesh.global and reviews.mesh.global.
+  # Optional
+  # Empty import allows to not subscribe FDS.
```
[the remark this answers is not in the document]
bartoszmajsak: We can have a following strategy in place:

- if `import` field does not exist it means we are not subscribing to any other remote
- if it's defined as empty (`{}`) that means we import from **ALL** remotes. That is common way of defining "all" cases, e.g. in `NetworkPolicy` having `podSelector: {}` means apply to all pods. See [`deny-all`](https://kubernetes.io/docs/concepts/services-networking/network-policies/#default-deny-all-ingress-traffic) and [`allow-all`](https://kubernetes.io/docs/concepts/services-networking/network-policies/#allow-all-ingress-traffic) as an example.
- if there are any remotes listed in the import list, only services from those will be imported.

docs/enhancements/1-crds-and-controllers.md
```diff
@@ -0,0 +1,430 @@
...
+      # Optional.
+      # Remote ingress type specifies how to manage client mTLS.
+      # Currently, only two types are supported: istio and openshift-router.
+      # If "openshift-router" is set the controller applies DestinationRules with SNI compatible with OpenShift Router.
+      # If "istio" is set client mTLS settings are not modified.
+      type: istio
+```
+
+The controller for this resource will be responsible only for updating the mesh configuration in other components,
+like Istio resource factory, FDS clients manager, etc. It will not manage any child resources.
```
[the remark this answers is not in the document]
bartoszmajsak: After a second thought I suggest we avoid going too deeply into implementation specifics, such as clients or resource factories. It's sufficient to state:

"This controller is responsible for updating the Federation's internal configuration and does not manage any Custom Resources."

docs/enhancements/1-crds-and-controllers.md
```diff
@@ -0,0 +1,430 @@
...
+                Owns(&v1.EnvoyFilter{}).
+                Watches(&corev1.Service{}, WithPredicates(checkIfMatchesExportRules)).
+                Watches(&corev1.Namespace{}, WithPredicates(checkIfMatchesExportRules)).
+                Watches(&MeshFederation{}).
+                Complete(r)
+}
+```
+
+All export-related resource will include owner reference pointing to `FederationServiceRules`,
+so deleting `FederationServiceRules` will result in removing these resources as well.
```
bartoszmajsak, line 282: I am afraid we are focusing too much on `ownerReference`s on its own, whereas the actual goal is to have a way of properly cleaning up related resources when given CR no longer exists. How it's done will emerge from the implementation.

docs/enhancements/1-crds-and-controllers.md
```diff
@@ -0,0 +1,430 @@
...
+> * FDS client can trigger the reconciliation loop sending an imported service via a channel to the controller.
+
+#### Egress gateway controller
+
+Egress gateway can be used in routing traffic to imported services, but it cannot be owned by `ImportedService`,
+because all imported services will be handled by one `Gateway` and one `VirtualService`, so this is N:1 relation.
+Therefore, resources related to egress gateway will be handled by a dedicated controller, and their owner will be `FederationServiceRules`.
+
+Responsibility of this controller:
+1. Watching `ImportedService` and reconciling `Gateway` and `VirtualService` if needed.
```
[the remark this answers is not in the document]
bartoszmajsak: In other words, what I think would be better to start with is to focus on what we need and why, not only to say which resources we reconcile. I am missing a bigger picture here.

docs/enhancements/1-crds-and-controllers.md
```diff
@@ -0,0 +1,430 @@
...
+        values:
+        - ratings
+```
+
+Responsibilities of the controller for this resource:
+1. Owning resources for exported services:
+   * `Gateway`;
+   * `ServiceEntry` - if custom search domain configured;
+   * `EnvoyFilter` - if the ingress type is "openshift-router".
+2. Watching services and namespaces and check if they match export rules.
```
bartoszmajsak, line 263: I think we should rephrase it - it's not about checking if they match our rules, but we are watching svc and ns that match defined rules and coordinates exporting between the meshes.

docs/enhancements/1-crds-and-controllers.md
```diff
@@ -0,0 +1,430 @@
...
+1. `MeshFederation` - cluster-scoped resource that includes general federation config, i.e. local settings, remote addresses and identities.
+2. `FederationServiceRules` - specifies rules for exporting and importing services; parent for export-related Istio resources, i.e. `Gateway`, `DestinationRule`, etc.; must be created by the mesh admin.
+3. `ImportedService` - represents an imported service; parent for import-related Istio resources, i.e. `ServiceEntry`, `WorkloadEntry`, etc.; managed only by the controller; should not be touch by users.
+
+All resources will contain [standard conditions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties) in their status.
+
+#### MeshFederation
+
+`MeshFederation` includes settings related only to mesh federation, like remote ingress address, identities, etc.
+This resource must be cluster-scoped, because it will be used as an owner for `ImportedService` resources, which can be created in multiple namespaces.
```
bartoszmajsak, line 36: If that is the only reason to have this resource as cluster-scoped, perhaps we can rethink how we deal with `ImportedService` and their related resources in the context of garbage collection - see https://github.com/openshift-service-mesh/federation/pull/120/files#r1913078586

docs/enhancements/1-crds-and-controllers.md
bartoszmajsak, line 1: I think it got lost in translation:

How about a bit more generous prefix for this file, i.e. `0001-crds-and-controllers.md`. This will help to see chronological order of design enhancements  :)

docs/enhancements/1-crds-and-controllers.md
```diff
@@ -0,0 +1,430 @@
...
+          # This is relevant only when the ingressType is openshift-router, but it cannot be empty.
+          name: tls-passthrough
+          # Optional.
+          # Port of the ingress gateway Service.
+          number: 15443
+    # Optional.
+    controlPlane:
+      # Optional.
+      # Control plane namespace used to create mesh-wide resources.
+      namespace: istio-system
```
bartoszmajsak, line 91: @jewertow If our controller will always be deployed in the same namespace as Istio is, I think we don't need this field at all. We can always get it from pod's env vars.

docs/enhancements/1-crds-and-controllers.md
```diff
@@ -0,0 +1,430 @@
...
+
+```yaml
+apiGroup: federation.openshift-service-mesh.io/v1alpha1
+kind: MeshFederation
+metadata:
+  name: default
+spec:
+  local:
+    # ID is a unique identifier of the FDS peer, and it's used as a suffix in its Service and ServiceAccount names.
+    id: east
```
bartoszmajsak, line 51: I think we should use `.metadata.name` instead. Then we offload handling the case of rename (remove and recreate child resources) to k8s clean-up instead, because user will deliberately need to first delete the resource and then create under the new name. That would imply that the existing RemoteMeshes are recreated by the user, but I would think of it as analogous case to remove Namespace with Services in it. If you want to rename the ns you do delete/create instead.

Im my opinion we should keep it simple for now and avoid corner cases unless they're really necessary from the UX perspective.

jewertow: OK

docs/enhancements/1-crds-and-kubebuilder.md
bartoszmajsak: *I mentioned this on slack a while back, but I think it's worth adding it here too:* If we want to trigger reconcile from some external flow (i.e. not a k8s event) we can use [`source.Channel`](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/source#Channel) for that.

PR #121, opened 2025-01-07
WIP: Introduce kubebuilder

eoinfennessy:


Makefile
```diff
@@ -136,30 +141,18 @@ $(PROTOBUF_GEN): $(PROTOBUF_API_SRC) $(PROTOC) $(PROTOC_GEN_GO) $(PROTOC_GEN_GRP
...
 		if ! grep -q "Licensed under the Apache License" $$file; then \
 			echo "Adding license to $$file"; \
 			cat $(LICENSE_FILE) $$file > temp && mv temp $$file; \
 		fi \
 	done
-	@rm -f $(LICENSE_FILE)
+
+.PHONY: generate
+generate: $(CONTROLLER_GEN) ## Generate code containing DeepCopy, DeepCopyInto, and DeepCopyObject method implementations.
+	$(CONTROLLER_GEN) object:headerFile=$(LICENSE_FILE) paths="./..."
```
bartoszmajsak, line 158: You don't need to have a phony target that is triggered each time, even when relevant files are not changed. Have a look hat ow `build` relies on `PROTOBUF_GEN` to determine if generation is needed. It might work here as well.

bartoszmajsak: @eoinfennessy one way to see which targets have been evaluated is to use `make build --debug=b `. You will see things like "Must remake target 'build'" in the logs.

PROJECT
```diff
@@ -0,0 +1,19 @@
...
+domain: federation.istio-ecosystem.io
+layout:
+- go.kubebuilder.io/v4
+projectName: federation
+repo: github.com/openshift-service-mesh/federation
+resources:
+- api:
+    crdVersion: v1
+  controller: true
+  domain: federation.istio-ecosystem.io
```
jewertow, line 14: I would change the domain to avoid duplication in the resource group:
```suggestion
  domain: istio-ecosystem.io
```

jewertow (commented):
How did you create `MeshFederation` resource? I was trying to add a resource with the following command:
```
kubebuilder create api --group federation --version v1alpha1 --kind ImportedService
```
but I got this error:
```
FATA failed to create API: unable to run pre-scaffold tasks of "base.go.kubebuilder.io/v4": cmd/main.go file should present in the root directory
```

Are you going to change `cmd` package in this PR?

eoinfennessy:
> How did you create MeshFederation resource?

I scaffolded a new project and API using kubebuilder, then copied the files in `api/v1alpha1` and pasted them into their current location.

> Are you going to change cmd package in this PR?

I wasn't intending to. I think enabling us to use `controller-gen` to generate code and CRDs is enough for this PR. The next step (a follow-up PR) will set up a simple controller for the `MeshFederation` resource. That work could include changing the project structure so that more APIs can be scaffolded using kubebuilder, but in my opinion this is not that important as creating each API is a one-time operation that can easily be achieved with a little copy-pasting, similar to what was done here.

jewertow:
Ok, that works for me.

PROJECT
eoinfennessy: Done. This has been changed from `federation.istio-ecosystem.io` to `openshift-serice-mesh.io`.

Makefile
```diff
@@ -123,43 +134,36 @@ $(PROTOC_GEN_DEEPCOPY):
...
 .PHONY: fix-imports
 fix-imports: $(GOIMPORTS) ## Fixes imports
 	$(GOIMPORTS) -local "github.com/openshift-service-mesh/federation" -w $(PROJECT_DIR)/
 
-LICENSE_FILE := /tmp/license.txt
+LICENSE_FILE := $(PROJECT_DIR)/hack/boilerplate.go.txt
 GO_FILES := $(shell find $(PROJECT_DIR)/ -name '*.go')
 
 .PHONY: add-license
 add-license: ## Adds license to all Golang files
```
jewertow, line 163: Do we need it if we have license in the boilerplate file?

jewertow (approved):

Makefile
eoinfennessy: Thanks. This has been addressed in commit f353806e1f9c3be0c7351b09d3cc3efbc19a4bc7 .

Makefile
bartoszmajsak: I don't think it covers all sources, only those which are generated (`object:headerFile`)

Makefile
```diff
@@ -13,12 +14,22 @@ PROTOC := $(LOCALBIN)/protoc
...
 PROTOBUF_GEN := $(shell find $(API_GEN_DIR) -type f -name "*.go")
 
+DEEP_COPY_DIR := $(PROJECT_DIR)/api/v1alpha1
+DEEP_COPY_SRC := $(shell find $(DEEP_COPY_DIR) -type f -not -name "*deepcopy.go")
+DEEP_COPY_GEN := $(DEEP_COPY_DIR)/zz_generated.deepcopy.go
+
+CRD_SRC_DIR := $(PROJECT_DIR)/api/v1alpha1
+CRD_SRC := $(shell find $(CRD_SRC_DIR) -type f -name "*.go")
+CRD_GEN_DIR := $(PROJECT_DIR)/chart/crds
+CRD_GEN := $(shell find $(CRD_GEN_DIR) -type f -name "*.yaml")
```
bartoszmajsak, line 31: I think we can simplify entire generation to one command, first we can remove `DEEP_COPY_*`

```suggestion
CRD_SRC_DIR := $(PROJECT_DIR)/api/v1alpha1
CRD_SRC := $(shell find $(CRD_SRC_DIR) -type f -name "*.go")
CRD_GEN_DIR := $(PROJECT_DIR)/chart/crds
CRD_GEN := $(shell find $(CRD_GEN_DIR) -type f -name "*.yaml")
```

Makefile
```diff
@@ -36,7 +47,7 @@ deps: ## Downloads required dependencies
 
 EXTRA_BUILD_ARGS?=
 .PHONY: build
-build: deps $(PROTOBUF_GEN) ## Builds the project
+build: deps $(PROTOBUF_GEN) $(DEEP_COPY_GEN) $(CRD_GEN) ## Builds the project
```
bartoszmajsak, line 50:
```suggestion
build: deps $(PROTOBUF_GEN) $(CRD_GEN) ## Builds the project
```

Makefile
bartoszmajsak, line 31:
```suggestion
CRD_SRC_DIR := $(PROJECT_DIR)/api/v1alpha1
CRD_SRC := $(shell find $(CRD_SRC_DIR) -type f -name "*.go")
CRD_GEN_DIR := $(PROJECT_DIR)/chart/crds
CRD_GEN := $(shell find $(CRD_GEN_DIR) -type f -name "*.yaml")
```

Makefile
```diff
@@ -123,43 +134,36 @@ $(PROTOC_GEN_DEEPCOPY):
...
 
 $(PROTOBUF_GEN): $(PROTOBUF_API_SRC) $(PROTOC) $(PROTOC_GEN_GO) $(PROTOC_GEN_GRPC) $(PROTOC_GEN_DEEPCOPY) ## Generates Go files from protobuf-based API files
 	@PATH=$(LOCALBIN):$$PATH $(PROTOC) --proto_path=$(PROTOBUF_API_DIR) --go_out=$(API_GEN_DIR) --go-grpc_out=$(API_GEN_DIR) --golang-deepcopy_out=:$(API_GEN_DIR) $(PROTOBUF_API_DIR)/**/*.proto
 
+$(DEEP_COPY_GEN): $(DEEP_COPY_SRC) ## Generate code containing DeepCopy, DeepCopyInto, and DeepCopyObject method implementations.
+	$(CONTROLLER_GEN) object:headerFile=$(LICENSE_FILE) paths="./..."
+
+$(CRD_GEN): $(CRD_SRC) ## Generate CustomResourceDefinition objects.
+	$(CONTROLLER_GEN) crd paths="./..." output:crd:artifacts:config=chart/crds
+
```
bartoszmajsak, line 154: Moving forward we will also add webhooks, so having the entire generation as one command will be more efficient.

What we missed here is dependency to the `controller-gen` itself - if it's not present in the path it will now be automatically fetched.

```suggestion
$(CRD_GEN): $(CRD_SRC) $(CONTROLLER_GEN) ## Generates CRDs and other controller-runtime scaffolding.
	$(CONTROLLER_GEN) paths="$(CRD_SRC_DIR)/..." \
		crd output:crd:artifacts:config="$(CRD_GEN_DIR)" \
		object:headerFile="$(LICENSE_FILE)"
```

bartoszmajsak (changes requested):
I think we can simplify the entire generation to one command. You can bundle my suggestions to a single commit.

Makefile
bartoszmajsak: There's more to that btw (RBAC generation, kube-builder directives in controllers etc), but we can fine-tune it as we go.

bartoszmajsak: One other thing from the top of my head - if someone opens PR with code changes but forgets to regenerate CRDs and other stuff we will end up with misaligned artifacts, i.e. YAMLs will be behind. I will capture it as an issue.

eoinfennessy: Done.

bartoszmajsak (approved):

bartoszmajsak:
Part of #148

PR #129, opened 2025-01-12
fix(examples): adjust helm values to recent API changes

jewertow:


bartoszmajsak (approved):
Sorry for missing it in my original PR, it was just too big :)

PR #131, opened 2025-01-16
feat(api): rename ExportedService to FederatedService and replace name and namespace with hostname

jewertow:
This change adjusts internal API to support service name aliases and custom search domains.


bartoszmajsak (approved):
LGTM. 

One concern is whether the multi-field composition of `metadata.name` might exceed its maximum length.

jewertow:
> One concern is whether the multi-field composition of metadata.name might exceed its maximum length.

253 characters is a pretty big limit.

PR #134, opened 2025-01-17
fix(fds): simplifies sync logic

bartoszmajsak:
My understanding of the current client-server flow is as follows:

- client connects to the server
- it requests push of predefined types (ExportedSvc) as initial push
  request - SotW
- on subscriber(client) connect, server does another SotW push

Not only we are sending the same data twice, but we also need to make
sure that handlers we define are aligned with initial push requests.
If we add new handler, we need to explictly define it as part of
initial push. It also needs to be defined (again) in onSubscribe
on the FDS server side.

This PR moves the initial sync responsibility to the client, ensuring
each handler automatically requests the data it needs. This avoids
redundant pushes and reduces code duplication.

### DIFF
--- a/cmd/federation-controller/main.go
+++ b/cmd/federation-controller/main.go
@@ -24,7 +24,6 @@ import (
 	"syscall"
 	"time"
 
-	discovery "github.com/envoyproxy/go-control-plane/envoy/service/discovery/v3"
 	routev1client "github.com/openshift/client-go/route/clientset/versioned"
 	istiokube "istio.io/istio/pkg/kube"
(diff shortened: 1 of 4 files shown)


eoinfennessy (approved):
Seems like a good change to me. @jewertow should probably also review this before adding okay-to-merge.

jewertow (approved):
`triggerFDSPushOnNewSubscription` is a weird leftover from the time when I was creating `adsc` and `adss` to support MCP-over-XDS and FDS. Thanks for this cleanup.

PR #136, opened 2025-01-20
feat(api): add crd and reconciler templates for FederatedService

jewertow:
I am adding these templates to avoid conflicts when we will work on different controllers in parallel.

How I tested this change:
```
keast apply -f - <<EOF
apiVersion: federation.openshift-service-mesh.io/v1alpha1
kind: FederatedService
metadata:
  name: test
  namespace: istio-system
spec:
  foo: bar
EOF
```
I got the following log:
```
2025-01-20T00:12:33Z    INFO    Reconciling object      {"controller": "federatedservice", "controllerGroup": "federation.openshift-service-mesh.io", "controllerKind": "FederatedService", "FederatedService": {"name":"test","namespace":"istio-system"}, "namespace": "istio-system", "name": "test", "reconcileID": "04eaf13a-6438-4428-807e-f2db9d69a521", "name": "test", "namespace": "istio-system"}
```


chart/templates/clusterrole.yaml
```diff
@@ -26,8 +26,8 @@ rules:
   verbs: ["get", "list", "create", "update", "patch", "delete"]
 {{- end }}
 - apiGroups: ["federation.openshift-service-mesh.io"]
-  resources: ["meshfederations"]
+  resources: ["meshfederations", "federatedservices"]
   verbs: ["create", "delete", "get", "list", "patch", "update", "watch"]
 - apiGroups: ["federation.openshift-service-mesh.io"]
-  resources: ["meshfederations/status"]
+  resources: ["meshfederations/status", "federatedservices/status"]
```
bartoszmajsak, line 32: @jewertow I wonder how this was added. By running `make` I can't see those changes, so I assume you did it by hand. 

Created #138 to follow-up on this.

eoinfennessy (commented):
Do we want/need any of the changes in `config`? I intentionally omitted these when adding the MeshFederation types because...
1. We are currently managing RBAC roles using helm.
2. Changes in `config/crds` do not really add anything because we are using helm for installation instead of kustomize.

I think the entire directory could be deleted. Alternatively, we could discuss generating RBAC types and/or using kustomize for installation. I could then create a PR to add back in any desired changes I previously omitted from `config`.

config/crd/kustomization.yaml
```diff
@@ -0,0 +1,20 @@
+# This kustomization.yaml is not intended to be run by itself,
+# since it depends on service name and namespace that are out of this kustomize package.
+# It should be run by config/default
+resources:
+- bases/federation.openshift-service-mesh.io_federatedservices.yaml
```
bartoszmajsak, line 5: As we are using Helm, we are not adhering to `kustomize` structure. This folder does not exist as we put our manifests elsewhere. We will have to investigate how to align generation of those manifests.

bartoszmajsak: Created #139 to track this work

bartoszmajsak (approved):
There are some gaps in the generation process which I discovered while playing with this PR. 

I created #137 and several sub-tasks to track the work in this area.

bartoszmajsak:
I think it's safe to merge to move forward with work, but we should address the concerns around the scaffolding in the nearest future.

jewertow:
@eoinfennessy that was a mistake. I don't know how did I miss those files...

PR #160, opened 2025-01-22
feat(build): replaces goimports with gci

bartoszmajsak:
While working on #159 my IDE moved local imports around
and `goimports` didn't reformat it properly. It would be good to have
clear rules for grouping and separating libraries we rely on that are
applied deterministically. It seems that `goimports` is fairly limited
and can handle only sorting and basic grouping. This can lead to noise
in PRs which we shouldn't bother about.

This PR changes `fix-imports` target to use `gci` tool instead. This
tool offers more granular control over group definitions so we can
ensure consistency across the codebase.

> [!NOTE]
> Even though `gci` is part of `golangci-lint` I deliberately added it
> as a standalone tool. Introducing the latter is a bigger task which I
> hope I can soon contribute.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,6 @@ CONTROLLER_TOOLS_VERSION ?= v0.16.4
 ## Required tooling.
 ## Needs to be defined early so that any target depending on given binary can resolve it when not present.
 LOCALBIN := $(PROJECT_DIR)/bin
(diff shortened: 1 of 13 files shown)


eoinfennessy (approved):
Thanks for introducing this!

PR #161, opened 2025-01-23
fix(api): changes scope of CRs to Namespaced

bartoszmajsak:
Current assumption is that we have all our resources namespaced.

This PR changes the scope of existing resources to be namespaced.

### DIFF
--- a/PROJECT
+++ b/PROJECT
@@ -10,6 +10,7 @@ repo: github.com/openshift-service-mesh/federation
 resources:
 - api:
     crdVersion: v1
+    namespaced: true
   controller: true
   domain: openshift-service-mesh.io
   group: federation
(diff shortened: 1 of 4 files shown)


jewertow (approved):

eoinfennessy (changes requested):
I think the PROJECT file may need to be updated to include `namespaced: true` similar to the `FederatedService` kind. I think controller-gen uses this field to determine if RBAC roles should be namespace-scoped or cluster-scoped.

```
resources:
- api:
    crdVersion: v1
    # add this line:
    namespaced: true
  controller: true
  domain: openshift-service-mesh.io
  group: federation
  kind: MeshFederation
  path: github.com/openshift-service-mesh/federation/api/v1alpha1
  version: v1alpha1
- api:
    crdVersion: v1
    namespaced: true
  controller: true
  domain: openshift-service-mesh.io
  group: federation
  kind: FederatedService
  path: github.com/openshift-service-mesh/federation/api/v1alpha1
  version: v1alpha1
```

eoinfennessy (approved):

PR #162, opened 2025-01-23
chore(build): reorganizes Makefile

bartoszmajsak:
As our build chain is growing it might be helpful to split Makefile to
several dedicated files which we can compose in the main `Makefile`.

This will help with identifying revelant targets quicker and group
together certain aspects of build process for easier maintenance.

This PR introduces `Makefile.tooling.mk`, where all targets for fetching
tools are now located.

### Additional fixes

- `make help` has a correct pattern matching and shows targets that
  contain numbers in them (i.e. `e2e`)
- test targets are grouped together
- `OUT_DIR` is prepended with `PROJECT_DIR`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,20 +1,15 @@
 PROJECT_DIR:=$(shell dirname $(abspath $(lastword $(MAKEFILE_LIST))))
-OUT_DIR:=out
+OUT_DIR:=$(PROJECT_DIR)/out
+
+.PHONY: default
+default: build add-license fix-imports test
 
 export ISTIO_VERSION ?= 1.23.0
-CONTROLLER_TOOLS_VERSION ?= v0.16.4
 
 ## Required tooling.
-## Needs to be defined early so that any target depending on given binary can resolve it when not present.
 LOCALBIN := $(PROJECT_DIR)/bin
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -97,57 +62,55 @@ e2e: kind-clusters ## Runs end-to-end tests against KinD clusters
...
+	$(CONTAINER_CLI) push $(HUB)/federation-controller:$(TAG)
 
-.PHONY: clean
-clean:
-	@rm -rf $(LOCALBIN) $(PROJECT_DIR)/$(OUT_DIR)
+.PHONY: docker
+docker: docker-build docker-push ## Combines build and push targets
 
-##@ Code Gen
+## Code Gen
```
eoinfennessy, line 94: Was removing `@` intentional?

```suggestion
##@ Code Gen
```

eoinfennessy (approved):
Looks good. I added one clarifying question/suggestion.

Makefile
bartoszmajsak: Yes, these are implicit targets which we always invoke. On top of that you won't know how to call them as they're all behind variables.

eoinfennessy: Ah yes! That makes sense. Thanks.

PR #164, opened 2025-01-24
feat(ctrl): encapsulates finalizer logic in reusable pkg

bartoszmajsak:
We are going to leverage finalizers to ensure we properly release
any resources owned by our controllers. That includes resources we
cannot garbage collect using `ownerRef` but also things like gRPC
connections we are going to establish between meshes.

Providing consistent approach to deal with this scenario helps us write cleaner and
maintainable code.

This change brings `finalizer` package to address this.

It has `finalizer.Handler` struct that:
- adds finalizer to a given object if it does not exists and then persist it
- performs cleanup defined as `CleanupFn` if the object is marked for deletion

Example usage in the controller reconcile loop:

```golang
finalizerHandler := finalizer.NewHandler(r.Client, "federation.openshift-service-mesh.io/mesh-federation")
if finalized, errFinalize := finalizerHandler.Finalize(ctx, meshFederation, func() error {
	// finalizer logic goes here
	return nil
}); finalized {
	return ctrl.Result{}, errFinalize
}

if finalizerAlreadyExists, errAdd := finalizerHandler.Add(ctx, meshFederation); !finalizerAlreadyExists {
	return ctrl.Result{}, errAdd
}
```

It seems to make sense to trigger another reconcile immediately after
adding finalizer, to ensure that the object the controller reconciles gets it
as soon as possible.

### Additional changes

Together with `finalizer` a set of common `retry` functions has been
added, so that `finalizer.Handler` can retry updating the object when
the conflict occurs.

### DIFF
--- a/internal/controller/finalizer/types.go
+++ b/internal/controller/finalizer/types.go
@@ -1,0 +1,98 @@
+// Copyright Red Hat, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the License);
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an AS IS BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package finalizer
+
+import (
+	"context"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+
+	"github.com/openshift-service-mesh/federation/internal/controller"
+)
+
+// FinalizeFn is a closure which can be used in the reconciler to define finalizer logic just before
+// the object is removed from kube-apiserver.
+type FinalizeFn func() error
+
+// Handler encapsulates finalizer handling. It can:
+// - add finalizer to a given object and persist it
+// - perform cleanup defined as FinalizeFn if the object is marked for deletion
+//
+// Example usage in the controller reconcile loop:
+//
+//	finalizerHandler := finalizer.NewHandler(r.Client, "federation.openshift-service-mesh.io/mesh-federation")
+//	if finalized, errFinalize := finalizerHandler.Finalize(ctx, meshFederation, func() error {
+//		// finalizer logic
+//		return nil
+//	}); finalized || errFinalize != nil {
+//		return ctrl.Result{}, errFinalize
+//	}
+//
+//	if justAdded, errAdd := finalizerHandler.Add(ctx, meshFederation); justAdded || errAdd != nil {
+//		return ctrl.Result{}, errAdd
+//	}
+type Handler struct {
+	cl            client.Client
+	finalizerName string
+}
+
+func NewHandler(cl client.Client, finalizerName string) *Handler {
+	return &Handler{
+		cl:            cl,
+		finalizerName: finalizerName,
+	}
+}
+
+// Add adds defined finalizer to the object and immediately persist it to ensure that finalizer has been added.
+// Returns true if finalizer was added and persisted in the cluster and error if the update failed.
+func (f *Handler) Add(ctx context.Context, obj client.Object) (bool, error) {
+	justAdded := controllerutil.AddFinalizer(obj, f.finalizerName)
+
+	if justAdded {
+		addFinalizer := func(saved client.Object) {
+			controllerutil.AddFinalizer(saved, f.finalizerName) // in case of conflict retry adding finalizer on the obj fetched from cluster
+		}
+		if _, errRetry := controller.RetryUpdate(ctx, f.cl, obj, addFinalizer); errRetry != nil {
+			return false, errRetry
+		}
+	}
+
+	return justAdded, nil
+}
+
+// Finalize executes finalizeFn when the object is about to be deleted.
+// Returns true if finalizer logic was successfully executed or error otherwise.
+func (f *Handler) Finalize(ctx context.Context, obj client.Object, finalizeFn FinalizeFn) (bool, error) {
+
+	shouldExecuteFinalize := !obj.GetDeletionTimestamp().IsZero() && controllerutil.ContainsFinalizer(obj, f.finalizerName)
+
+	if shouldExecuteFinalize {
+		if err := finalizeFn(); err != nil {
+			return false, err
+		}
+
+		removeFinalizer := func(saved client.Object) {
+			controllerutil.RemoveFinalizer(saved, f.finalizerName) // in case of conflict retry removing finalizer on the obj fetched from cluster
+		}
+		if _, errRetry := controller.RetryUpdate(ctx, f.cl, obj, removeFinalizer); errRetry != nil {
+			return false, errRetry
+		}
+
+		return true, nil
+	}
+
+	return false, nil
+}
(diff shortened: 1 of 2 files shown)


internal/controller/finalizer/types.go
```diff
@@ -0,0 +1,98 @@
...
+		}
+		if _, errRetry := controller.RetryUpdate(ctx, f.cl, obj, removeFinalizer); errRetry != nil {
+			return false, errRetry
+		}
+
+		return true, nil
+	}
+
+	return false, nil
+}
```
eoinfennessy, line 98: I think the method should return true if the object should be finalized.

Finalizer success is then indicated by the absence of an error, and only the absence of an error.

Usage after such a change is simplified:
```
if finalizeNecessary, err := finalizerHandler.Finalize(ctx, meshFederation, func() error {
    // finalizer logic
    return nil
}); finalizeNecessary {
    return ctrl.Result{}, err
}
```

```suggestion
// Finalize executes finalizeFn when the object is about to be deleted.
// Returns true if the object should be finalized.
// Returns an error if the finalize function was unsuccessful or the object update failed.
func (f *Handler) Finalize(ctx context.Context, obj client.Object, finalizeFn FinalizeFn) (bool, error) {

	shouldExecuteFinalize := !obj.GetDeletionTimestamp().IsZero() && controllerutil.ContainsFinalizer(obj, f.finalizerName)

	if shouldExecuteFinalize {
		if err := finalizeFn(); err != nil {
			return true, err
		}

		removeFinalizer := func(saved client.Object) {
			controllerutil.RemoveFinalizer(saved, f.finalizerName) // in case of conflict retry removing finalizer on the obj fetched from cluster
		}
		if _, errRetry := controller.RetryUpdate(ctx, f.cl, obj, removeFinalizer); errRetry != nil {
			return true, errRetry
		}

		return true, nil
	}

	return false, nil
}
```

internal/controller/finalizer/types.go
```diff
@@ -0,0 +1,98 @@
...
+		addFinalizer := func(saved client.Object) {
+			controllerutil.AddFinalizer(saved, f.finalizerName) // in case of conflict retry adding finalizer on the obj fetched from cluster
+		}
+		if _, errRetry := controller.RetryUpdate(ctx, f.cl, obj, addFinalizer); errRetry != nil {
+			return false, errRetry
+		}
+	}
+
+	return justAdded, nil
+}
```
eoinfennessy, line 74: Similar to the `Finalize` method, I think using the boolean to indicate whether or not the finalizer was already present could make the usage simpler:

```
if finalizerPresent, err := finalizerHandler.Add(ctx, meshFederation); !finalizerPresent {
	return ctrl.Result{}, err
}
```

```suggestion
// Add adds the defined finalizer to the object and then persists it if the finalizer was not already present.
// Returns true if the finalizer was already present.
// Returns an error if updating the object failed.
func (f *Handler) Add(ctx context.Context, obj client.Object) (bool, error) {
	justAdded := controllerutil.AddFinalizer(obj, f.finalizerName)

	if justAdded {
		addFinalizer := func(saved client.Object) {
			controllerutil.AddFinalizer(saved, f.finalizerName) // in case of conflict retry adding finalizer on the obj fetched from cluster
		}
		_, errRetry := controller.RetryUpdate(ctx, f.cl, obj, addFinalizer)
		return false, errRetry
	}

	return true, nil
}
```

internal/controller/finalizer/types.go
```diff
@@ -0,0 +1,98 @@
...
+//	if finalized, errFinalize := finalizerHandler.Finalize(ctx, meshFederation, func() error {
+//		// finalizer logic
+//		return nil
+//	}); finalized || errFinalize != nil {
+//		return ctrl.Result{}, errFinalize
+//	}
+//
+//	if justAdded, errAdd := finalizerHandler.Add(ctx, meshFederation); justAdded || errAdd != nil {
+//		return ctrl.Result{}, errAdd
+//	}
```
eoinfennessy, line 46: Change related to other two comments.

```suggestion
//	finalizerHandler := finalizer.NewHandler(r.Client, "federation.openshift-service-mesh.io/mesh-federation")
//	if finalizeNecessary, errFinalize := finalizerHandler.Finalize(ctx, meshFederation, func() error {
//		// finalizer logic
//		return nil
//	}); finalizeNecessary {
//		return ctrl.Result{}, errFinalize
//	}
//
//	if finalizerPresent, err := finalizerHandler.Add(ctx, meshFederation); !finalizerPresent {
//		return ctrl.Result{}, errAdd
//	}
```

internal/controller/finalizer/types.go
bartoszmajsak: The intention of this boolean is to distinguish if it's the first time we added the finalizer (so we updated the state in the cluster) . Therefore we need to return `justAdded` at the end to know if we should finish and wait for another reconcile or not. If we keep returning `true` we will never execute anything in reconcile which comes after piece of logic.

We need to make sure however that our predicates watch on finalizer changes (`predicate.GenerationChangedPredicate` does not, so we need one extra if we use it)

bartoszmajsak: Maybe it should be named `FinalizeIfNeeded`

internal/controller/finalizer/types.go
eoinfennessy: > Returns true if the object should be finalized.

I should have made this clearer. In my suggestion this boolean is used to indicate that finalize was attempted. Here's an updated suggestion, rephrased to use an early return too:

```suggestion
// Finalize executes finalizeFn when the object is about to be deleted.
// Returns true if the execution of the finalizeFn plus object update were attempted.
// Returns an error if the finalize function was unsuccessful or the object update failed.
func (f *Handler) Finalize(ctx context.Context, obj client.Object, finalizeFn FinalizeFn) (bool, error) {
	if shouldExecuteFinalize := !obj.GetDeletionTimestamp().IsZero() && controllerutil.ContainsFinalizer(obj, f.finalizerName); !shouldExecuteFinalize {
		return false, nil
	}

	if err := finalizeFn(); err != nil {
		return true, err
	}
	removeFinalizer := func(saved client.Object) {
		controllerutil.RemoveFinalizer(saved, f.finalizerName) // in case of conflict retry removing finalizer on the obj fetched from cluster
	}
	if _, errRetry := controller.RetryUpdate(ctx, f.cl, obj, removeFinalizer); errRetry != nil {
		return true, errRetry
	}
	return true, nil
}
```

Usage:
```
if finalizeAttempted, err := finalizerHandler.Finalize(ctx, meshFederation, func() error {
    // finalizer logic
    return nil
}); finalizeAttempted {
    return ctrl.Result{}, err
}
```

internal/controller/finalizer/types.go
bartoszmajsak: Ah ok, I missed that in the diff! That's in fact better!

I wonder `EnsureAdded` or something similar wouldn't be more helpful name. WDYT?

eoinfennessy: > I wonder EnsureAdded or something similar wouldn't be more helpful name. WDYT?

Yeah, how about something like `AddIfNeeded` and `FinalizeIfNeeded`?

internal/controller/finalizer/types.go
```diff
@@ -60,31 +60,32 @@ func NewHandler(cl client.Client, finalizerName string) *Handler {
...
+	})
+
 	return false, errRetry
 }
 
-// Finalize executes finalizeFn when the object is about to be deleted.
-// Returns true if the execution of the finalizeFn plus object update were attempted.
-// Returns an error if the finalize function was unsuccessful or the object update failed.
+// Finalize executes cleanup logic defined in finalizeFn only if the object is marked for deletion.
+// Returns true if finalizer logic was performed successfully.
```
eoinfennessy, line 75: This is not accurate. `true` indicates that the finalizer logic plus update was _attempted_, not that it was performed successfully.

eoinfennessy:
```suggestion
// Returns true if finalizer logic plus object update was attempted.
```

eoinfennessy (approved):

PR #165, opened 2025-01-24
feat(ctrl): predicate on finalizer change

bartoszmajsak:
`finalizer.Handler` introduced in #164 is expected to retrigger
reconcile when finalizer is added for the first time. This ensures that
given controller adds its cleanup hook to the resource as soon as
possible.

However adding finalizer does not increase generation of the resource,
therefore using `WithEventFilter(predicate.GenerationChangedPredicate{})`
will not cause another reconcile.

This PR brings `FinalizerChanged` predicate which can be combined:

```golang
WithEventFilter(
	predicate.Or(
		predicate.GenerationChangedPredicate{},
		FinalizerChanged(),
	),
).
```

and it will trigger reconciliation on the finalizer change.

### DIFF
--- a/internal/controller/meshfederation/meshfederation_controller.go
+++ b/internal/controller/meshfederation/meshfederation_controller.go
@@ -20,8 +20,10 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
 
 	"github.com/openshift-service-mesh/federation/api/v1alpha1"
+	"github.com/openshift-service-mesh/federation/internal/controller"
 )
 
 // +kubebuilder:rbac:groups=federation.openshift-service-mesh.io,resources=meshfederations,verbs=get;list;watch;create;update;patch;delete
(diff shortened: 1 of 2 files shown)


eoinfennessy:
I'm kind of leaning towards the opinion that we should not use `predicate.GenerationChangedPredicate`.

This would mean more reconcile requests (one additional request every time we update the status), but would ensure that the status remains consistent with the observed state when, for instance, a user accidentally updates/deletes a MeshFederation's status.

Is an additional reconcile request really that expensive if the state of a MeshFederation has previously been reconciled?

bartoszmajsak:
As long as we don't have any external system writing to our status or some sort of self-healing mechanism I would start by ignoring changes to `.status` field.

bartoszmajsak:
> Is an additional reconcile request a performance issue if the state of a MeshFederation has previously been reconciled (or it has reached an irreconcilable state)?

With the current line of thinking each reconcile will trigger SotW push to connected clients, so I would rather minimize amount of such calls :)

eoinfennessy:
> With the current line of thinking each reconcile of MeshFederation will trigger SotW push to connected clients, so I would rather minimize amount of such calls :)

Ah, okay! Adding this predicate makes sense I think. Thanks!

eoinfennessy (approved):

PR #167, opened 2025-01-24
feat(testing): introduces k8s envtest skeleton

### LINKED ISSUE #146: Define testing strategy for Kubernetes Reconcilers
We need to establish a clear testing strategy to ensure the Kubernetes reconciler logic behaves as expected. 

#### Goals

- Fast feedback loop during development of the controllers.
- Test skeleton to be followed across all controllers.

We can also document our approach in ADR.

bartoszmajsak:
This PR provides a skeleton for testing controller using K8s EnvTest
testing harness. This approach enables us to run reasonably fast
integration tests interacting with kube-apiserver.

This way we can have automated feedback loop for verifying parts of the
system behaviour without a need to run entire e2e suite. This is
especially handy right now, as we are fleshing out controller-runtime
implementation and have e2e tests for the legacy implementation.

Besides simple test this commit brings a facade that simplifies
configuration and use Kubernetes EnvTest [1] library. The main goal of
this enhancement is to reduce boilerplate of test suite
bootstrap code by hiding the details of EnvTest setup and
wiring the controller-under-test.

Additionally, the facade struct gives an unified way of manipulating
cluster under test as it is composed of k8s client.Client and custom
Cleaner (see its godoc for details why it is needed) struct.

The tests are written in Ginkgo [2] and use Gomega matchers [3].
These tools are used in many k8s projects, especially controllers, as
`controller-runtime` generates basic template for tests based on those
libraries. Ginkgo provides a structured and expressive way to write tests.

### Extras

- automates fetching of external CRDs used when testing. This way we can always have all required CRDs installed in the cluster before starting tests. 
- simplifies scheme registration

> [!NOTE] 
> I’m open to adjusting the testing approach if we find this one isn't ideal. 
> This PR is an iteration based on my previous work in projects where 
> controllers were used.

[1] https://book.kubebuilder.io/reference/envtest
[2] https://onsi.github.io/ginkgo
[3] https://onsi.github.io/gomega/

Fixes #146


internal/controller/meshfederation/meshfederation_controller.go
```diff
@@ -33,18 +40,55 @@ type Reconciler struct {
...
+
 	return ctrl.Result{}, nil
 }
 
 // SetupWithManager sets up the controller with the Manager.
 func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
+		Named("mesh-federation-ctrl").
 		For(&v1alpha1.MeshFederation{}).
+		WithEventFilter(predicate.GenerationChangedPredicate{}).
```
bartoszmajsak, line 92: FWIW There's a discussion about filtering out `.status` changes from reconciliation in #165.

bartoszmajsak:
`+15,796` is b/c of Istio CRD  YAML. I thought about excluding it, as it's anyway fetched every time we run tests, but I then wondered if we shouldn't keep it b/c of offline builds.

Happy to remove it.

jewertow (approved):

PR #171, opened 2025-01-27
docs: initial architectural diagram 

bartoszmajsak:
Based on the design discussions I updated the high-level diagram
describing our current approach on how reconciliation should work.

It describes responsibilities of each controller as well as
relationships between resources we manage.

Besides "diagram-as-code" this PR brings auto-generated PNGs based on
latest changes and a simple script `export.sh` which takes care of it.

It can be later used to automate as GitHub Action to keep images in sync
with `.drawio` changes.


eoinfennessy (approved):
Great to have this! I verified `export.sh` s working as expected.

jewertow (approved):

PR #180, opened 2025-03-13
docs: update samples for OpenShift

jewertow:
Samples for OpenShift worked only with the operator 3.0.0-tp.2, which is no longer available.


bartoszmajsak (approved):
Thanks!

PR #181, opened 2025-03-17
docs: replace `trustDomainAliases` with `caCertificates[].trustDomains` in examples for SPIRE

jewertow:
Using `trustDomainAliases` is not the right solution when federated services from different trust domains may have overlapping service account and namespace names. Then authorization policies defined for a single trust domain are automatically generated for all trust domain aliases, and this may be not desired in multi-mesh deployments, so we should avoid using this solution.

We also use `trustDomainAliases` in e2e tests, but that can be fixed in a follow-up.


bartoszmajsak (approved):
LGTM. I can take a stab at the follow-up if you haven't already.

jewertow:
> I can take a stab at the follow-up if you haven't already

Go ahead :)

